PR #5, opened 2017-06-15
http-fetcher with basic tests

evanshauser:
This include the most basic fetcher, http-fetcher with a couple of tests. There are a couple constructs that illustrate possible additions to the specification and a bunch of notes for additions that need to be made to make this http-fetcher a canonical example.


src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export interface FetchResult {
+  data: any;
+  errors?: any;
+  extensions?: any;
+  context?: object;
+}
+
+export type UnsubscribeHandler = () => void;
```
helfer: I think this is now different in the TC39 proposal. Do we want to try to align here?

tests/index.ts
```diff
@@ -1,9 +1,115 @@
...
+    const fetcher = new HttpFetcher('', fetch);
+    const observable = fetcher.request({
+      query: sampleQuery,
+      operationName: 'SampleQuery',
+    });
+    //observableToPromise return a promise
+    observable.subscribe({
+      next,
+      error: (error) => assert(false),
+      complete: () => { assert(next.calledOnce); done(); },
```
helfer: Have you tried if this test can fail at all?

tests/index.ts
```diff
@@ -1,9 +1,115 @@
...
+describe('HttpFetcher', () => {
+  it('does not need any constructor arguments', () => {
+    assert.doesNotThrow(() => new HttpFetcher());
+  });
+
+  it('calls next and then complete', (done) => {
+    const next = sinon.spy();
+    const data = {hello: 'world'};
+    fetchMock.get('*', data);
+    const fetcher = new HttpFetcher('', fetch);
```
helfer: We should make the `fetch` argument optional. We'll add a separate test for adding the argument later, which can use `sinon.spy` to check that the function is getting called.

src/link.ts
```diff
@@ -0,0 +1,73 @@
...
+        observable.subscribe({
+          next: resolve,
+          error: reject,
+        });
+      });
+    };
+  }
+
+  private static buildLinkChain(links: ApolloLink[]): NextLink {
+    const _links = [...links];
```
evanshauser: This way of creating the connections between the links makes a new array for each request. It mirrors the middleware and afterware code from the netowrk interface of `apollo-client`(also in `apollo-fetch`) the `next` function on demand. We could create a closure and keep an index that we increment

src/setContextLink.ts
```diff
@@ -0,0 +1,29 @@
...
+  ApolloLink,
+} from './link';
+
+export default class SetContextLink extends ApolloLink {
+
+  constructor (private context?: any) {
+    super();
+  }
+
+  public request(operation: Operation, forward: NextLink): Observable<FetchResult> {
```
jbaxleyiii: @evanshauser I'm concerned about the current `forward` pattern since it could be optional? I'm a little confused at forward and how we even do things like enforcing passing a forward into a forward?

src/link.ts
```diff
@@ -0,0 +1,182 @@
...
+export function execute(link: ApolloLink, operation: GraphQLRequest): Observable<FetchResult> {
+  validateOperation(operation);
+
+  if (operation.context === undefined) {
+    operation.context = {};
+  }
+  if (operation.variables === undefined) {
+    operation.variables = {};
+  }
+  if (operation.query === undefined) {
```
jbaxleyiii: what is this for?

src/link.ts
```diff
@@ -0,0 +1,182 @@
...
+          types {
+            name
+          }
+        }
+      }
+    `;
+  }
+  const _operation = transformOperation(operation);
+
+  return link.request(_operation) || Observable.of();
```
jbaxleyiii: should there be a `forward` passed to this request?

src/cacheLink.ts
```diff
@@ -0,0 +1,44 @@
+import {
```
stubailo: This file is unused - can we move it into the tests, or an unused folder, or remove it?

stubailo: This file is unused - can we move it into the tests, or an unused folder, or remove it?

src/httpLink.ts
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    this.requestLink = new SingleRequestLink({
+      ...(fetchParams || {}),
+      fetch: this._fetch,
+    });
+  }
+
+  public request(operation: Operation): Observable<FetchResult> | null {
+    this.headers = operation.context && operation.context.headers || {};
```
stubailo: For every link, we should make sure to document what it expects on the context.

src/httpUtils.ts
```diff
@@ -0,0 +1,46 @@
+import {
```
stubailo: This file is not used. Let's remove it?

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+  validateOperation(operation);
+
+  if (!operation.context) {
+    operation.context = {};
+  }
+  if (!operation.variables) {
+    operation.variables = {};
+  }
+  if (!operation.query) {
+    console.warn(`query should either be a sting or GraphQL AST`);
```
stubailo: `sting` -> `string`

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+    operation.query = <DocumentNode>{};
+  }
+
+  const _operation: Operation = transformOperation(operation);
```
stubailo: Maybe `transformedOperation`? `_operation` isn't a super descriptive name.

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+  return link.request(_operation) || Observable.of();
+}
+
+const toPromise = (link: ApolloLink) => {
+  return (operation: Operation, forward?: NextLink) => {
+    const observable = link.request(operation, forward);
+
+    return new Promise((resolve, reject) => {
+      observable.subscribe({
+        next: resolve,
```
stubailo: I think if the observable has multiple results this should print a warning

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+    link = new FunctionLink(link);
+  }
+  return {
+    request: toPromise(link),
+  };
+}
+
+function transformOperation(operation) {
+  if (typeof operation.query === 'string') {
+    operation = {
```
stubailo: It's odd that this is taking pains not to mutate the input, but the below code doesn't.

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+      ...operation,
+      query: parse(operation.query),
+    };
+  }
+
+  if (!operation.operationName) {
+    operation.operationName = operation.query.definitions &&
```
stubailo: This will fail on:

```graphql
fragment aFragmentDefinition on Query { lol }

query theQuery { ...lol }
```

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+      operation.query.definitions.length &&
+      operation.query.definitions[0].name || '';
+  }
+
+  return operation;
+}
+
+export class TerminatedLink extends ApolloLink {
```
stubailo: Maybe `TerminatingLink`

stubailo: Also, I don't think this class is used right now, other than in a ternary expression where it doesn't provide any type safety.

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+  constructor(private f: RequestHandler) {
+    super();
+  }
+
+  public request(operation: Operation): Observable<FetchResult> {
+    return this.f(operation) || Observable.of();
+  }
+}
+
+export class TerminatedConcat extends ApolloLink {
```
stubailo: Same as above, this doesn't provide any extra type safety where it's used

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+    if (ApolloLink.isTerminating(left) && ApolloLink.isTerminating(right)) {
+      return new TerminatedSplit(test, left, right);
+    }
+
+    return new SplitLink(test, left, right);
+  }
+
+  public static toLink(link: ApolloLink | RequestHandler): ApolloLink {
+    if (typeof link === 'function') {
+      return link.length <= 1 ? new TerminatedLink(link) : new FunctionLink(link);
```
stubailo: If we want to keep these here, we should use `.isTerminating`. But I don't think this helps since the function returns an `ApolloLink` anyway.

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+    right: ApolloLink | RequestHandler = ApolloLink.passthrough(),
+  ): ApolloLink {
+    return this.concat(<ApolloLink>ApolloLink.split(test, left, right));
+  }
+
+  // join two Links together
+  public concat(link: ApolloLink | RequestHandler): ApolloLink {
+    if (this.request.length <= 1) {
+      const warning = Object.assign(
+        new Error(`You are calling concat a terminating link, which will have no effect`),
```
stubailo: `calling concat a` -> `calling concat on a`

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+    left: ApolloLink | RequestHandler,
+    right: ApolloLink | RequestHandler = ApolloLink.passthrough(),
+  ): ApolloLink {
+    return this.concat(<ApolloLink>ApolloLink.split(test, left, right));
+  }
+
+  // join two Links together
+  public concat(link: ApolloLink | RequestHandler): ApolloLink {
+    if (this.request.length <= 1) {
+      const warning = Object.assign(
```
stubailo: What is this assign for?

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+        next: resolve,
+        error: reject,
+      });
+    });
+  };
+};
+
+export function asPromiseWrapper(link: ApolloLink | RequestHandler) {
+  if (typeof link === 'function') {
+    link = new FunctionLink(link);
```
stubailo: Can't we use `toLink` here?

src/linkUtils.ts
```diff
@@ -0,0 +1,41 @@
+import {
```
stubailo: Should we move `toLink`, `isTerminating`, etc, here?

src/mockLink.ts
```diff
@@ -0,0 +1,22 @@
+import {
```
stubailo: This is not used

src/pollingLink.ts
```diff
@@ -0,0 +1,49 @@
...
+
+
+export default class PollingLink extends ApolloLink {
+
+  private pollInterval: number;
+  private timer;
+
+  constructor(pollInterval?: number) {
+    super();
+    this.pollInterval = pollInterval || 0;
```
stubailo: I'd say we should just throw if the poll interval is not set. `0` doesn't seem like a reasonable default.

src/pollingLink.ts
```diff
@@ -0,0 +1,49 @@
...
+      const subscriber = {
+          next: (data) => {
+            observer.next(data);
+          },
+          error: (error) => observer.error(error),
+      };
+
+      const poll = (() => {
+        forward(operation).subscribe(subscriber);
+      }).bind(this);
```
stubailo: I don't think the bind here is necessary.

src/retryLink.ts
```diff
@@ -0,0 +1,66 @@
...
+            }, this.interval(this.delay, this.count));
+          } else {
+            observer.error(error);
+          }
+        },
+        complete: () => {
+          observer.complete();
+        },
+        // This causes an error, not sure why
+        // complete: observer.complete
```
stubailo: You probably need `observer.complete.bind(observer)`.

src/pollingLink.ts
evanshauser: should be interval

src/setContextLink.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (!operation.context) {
+      operation.context = {};
+    }
+    operation.context = {
```
evanshauser: This should be a recursive merge, since a link could overwrite an entire object, such as headers

src/linkUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+  const OPERATION_FIELDS = ['query', 'operationName', 'variables', 'context'];
+  for (let key of Object.keys(operation)) {
+    if (OPERATION_FIELDS.indexOf(key) < 0) {
+      throw new Error(`illegal argument: ${key}`);
+    }
+  }
+}
+
+export function ensureForward(forward?: NextLink): void {
+  if (!forward || typeof forward !== 'function') {
```
jbaxleyiii: the type check here should be enough https://runkit.com/jbaxleyiii/596d174b03c14800141f7fcf

src/pollingLink.ts
jbaxleyiii: @evanshauser are you open to it being a function that given the operation, can determine the interval?

src/pollingLink.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      if (this.pollInterval) {
+        this.timer = setInterval(poll, this.pollInterval);
+      }
+
+      forward(operation).subscribe(subscriber);
+
+      return () => {
+        clearInterval(this.timer);
```
jbaxleyiii: I think you would also need to clean up the subscribers that have been created during polling yes? Since each interval will create a new subscriber

jbaxleyiii: also, we need to check to make sure `this.timer` exists. I would probably just use a local variable instead of attaching it to the instance as well.

src/retryLink.ts
```diff
@@ -0,0 +1,66 @@
...
+    interval?: (delay: number, count: number) => number,
+  }) {
+    super();
+    this.max = params && params.max || 10;
+    this.delay =  params && params.delay || 300;
+    this.interval =  params && params.interval || this.defaultInterval;
+  }
+
+  public request(operation: Operation, forward: NextLink): Observable<FetchResult> {
+    ensureForward(forward);
```
jbaxleyiii: should this be attached to the Link interface so it can be used without importing? The forward need is still I think the confusingish part here, so lets try to make it an easy thing for library developers to use

src/retryLink.ts
```diff
@@ -0,0 +1,66 @@
...
+
+
+    return new Observable(observer => {
+      const subscriber = {
+        next: data => {
+          observer.next(data);
+          this.count = 0;
```
jbaxleyiii: I would switch these in case observer.next throws, we wan't the count reset anytime `.next` is called

src/retryLink.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+        error: error => {
+          this.count++;
+          if (this.count < this.max) {
+            setTimeout(() => {
```
jbaxleyiii: lets store the timer so it can be cleaned up if it needs to be

src/retryLink.ts
```diff
@@ -0,0 +1,66 @@
...
+          }
+        },
+        complete: () => {
+          observer.complete();
+        },
+        // This causes an error, not sure why
+        // complete: observer.complete
+      };
+
+      forward(operation).subscribe(subscriber);
```
jbaxleyiii: We should return an unsubscriber here that cleans up subscriptions and timers

src/retryLink.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+  }
+
+  private defaultInterval = (delay, count) => delay;
```
jbaxleyiii: we don't need to access `count` if not being used

src/setContextLink.ts
jbaxleyiii: fwiw, recursive merges are really expensive :( I would say that this link overwrites and previous context of the same keys when being added.

@evanshauser when we call request within the `ApolloLink` interface, thoughts on creating a copy of the operation and passing it down?

src/singleRequestLink.ts
```diff
@@ -0,0 +1,51 @@
...
+  createApolloFetch,
+  ApolloFetch,
+} from 'apollo-fetch';
+import { print } from 'graphql/language/printer';
+
+import {
+  ApolloLink,
+} from './link';
+
+export default class SingleRequestLink extends ApolloLink {
```
jbaxleyiii: how does this differ from `httpLink`? Why would I use this over that?

src/singleRequestLink.ts
```diff
@@ -0,0 +1,51 @@
...
+    fetch?: ApolloFetch,
+  }) {
+    super();
+    this._fetch = fetchParams && fetchParams.fetch || createApolloFetch({ uri: fetchParams && fetchParams.uri });
+  }
+
+  public request(operation: Operation): Observable<FetchResult> {
+    const request = {
+      ...operation,
+      query: print(operation.query),
```
jbaxleyiii: this isn't needed since execute does it?

src/singleRequestLink.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    return new Observable<FetchResult>(observer => {
+      this._fetch(request)
+      .then(data => {
```
jbaxleyiii: *nit spacing here is off

src/types.ts
```diff
@@ -0,0 +1,49 @@
+import { DocumentNode } from 'graphql/language/ast';
+import * as Observable from 'zen-observable';
+
+export interface GraphQLRequest {
+  //This string could be an id for a persisted query
```
jbaxleyiii: lets remove this comment as I don' think it is correct

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  query?: string | DocumentNode;
+  variables?: object;
```
jbaxleyiii: lets make this and context [a `Record` type](https://www.typescriptlang.org/play/index.html#src=let%20x%3A%20Record%3Cstring%2C%20any%3E%0D%0A%0D%0Ax%20%3D%20%7B%20a%3A%201%20%7D%0D%0Ax.b%20%3D%204%0D%0A%0D%0Alet%20y%3A%20Object%0D%0Ay%20%3D%20%7B%20a%3A%201%20%7D%0D%0Ay.b%20%3D%204)

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  context?: object;
+}
+
+export interface Link {
```
jbaxleyiii: Is this used?

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  request (operation: Operation): Promise<FetchResult>;
+}
+
+export interface Operation {
+  query: DocumentNode;
+  variables?: {
+    [variables: string]: any;
+  };
+  operationName?: string;
+  context?: {
```
jbaxleyiii: same comment for Record as above

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export interface FetchResult {
+  data: any;
+  errors?: any;
+  extensions?: any;
+  context?: object;
+}
+
+export interface Subscription {
```
jbaxleyiii: Lets check to see if `@types/zen-observable` has these so we can use the types of the underlying observable implementation

tests/httpLink.ts
```diff
@@ -0,0 +1,215 @@
...
+  });
+
+  afterEach(() => {
+    fetchMock.restore();
+  });
+
+  it('raises warning if called with concat', () => {
+    const link = ApolloLink.from([new HttpLink()]);
+    const _warn = console.warn;
+    console.warn = warning => assert.property(warning, 'message');
```
jbaxleyiii: its been a while since I've used mocha, will this error if console.warn isn't called?

tests/httpLink.ts
```diff
@@ -0,0 +1,215 @@
...
+    assert.deepEqual(link.concat((operation, forward) => forward(operation)), link);
+    console.warn = _warn;
+  });
+
+  it('does not need any constructor arguments', () => {
+    assert.doesNotThrow(() => new HttpLink());
```
jbaxleyiii: shouldn't it? What would it do in that case?

tests/httpLink.ts
```diff
@@ -0,0 +1,215 @@
...
+  });
+
+  it('calls next and then complete', (done) => {
+    const next = sinon.spy();
+    const link = new HttpLink({uri: 'data'});
+    const observable = link.request({
```
jbaxleyiii: For all of these tests, I think we should be using `execute(link, operation)` instead of the `.request`. As link grows, I can see things being done during execution that each individual ink wouldn't need to ensure in its request (for instance mutation of the operation). 

I don't think `request` is really meant to be used publically on the link, instead it is used by the library itself to build the execution chain

tests/httpLink.ts
```diff
@@ -0,0 +1,215 @@
...
+      query: sampleMutation,
+      context,
+      variables,
+    });
+    observable.subscribe({
+      next,
+      error: (error) => assert(false),
+      complete: () => {
+        const body = JSON.parse(fetchMock.lastCall()[1]['body']);
+        assert.equal(body['query'], print(sampleMutation));
```
jbaxleyiii: any reason for not using dot notation here?

tests/link.ts
```diff
@@ -0,0 +1,862 @@
...
+    id
+  }
+}`;
+
+describe('ApolloLink(abstract class)', () => {
+
+  describe('concat', () => {
+    it('should concat a function', (done) => {
+      const returnOne = new SetContextLink({ add: 1 });
+      const link = returnOne.concat((operation, forward) => Observable.of({ data: operation.context.add }));
```
jbaxleyiii: @evanshauser this shows my potential fear with forward :(

I'm assuming the library would not see this as a terminating link, yet forward is not used so it in essence is terminating. I think with copy paste / boilerplate, this could really get people confused

tests/link.ts
```diff
@@ -0,0 +1,862 @@
...
+        operation = {
+          ...operation,
+          context: {
+            add: operation.context.add + 2,
+          },
+        };
+        return forward(operation);
+      });
+      const mock2 = new MockLink((op, forward) => Observable.of({ data: op.context.add + 2 }));
+      const mock3 = new MockLink((op, forward) => Observable.of({ data: op.context.add + 3 }));
```
jbaxleyiii: lets move `MockLink` to the `/tests` folder so it won't be confused as an external link

tests/singleRequestLink.ts
```diff
@@ -0,0 +1,9 @@
+import { assert } from 'chai';
+import SingleRequestLink from '../src/singleRequestLink';
+
+
+describe('SingleRequestLink', () => {
```
jbaxleyiii: I don't think this is needed since the link won't be shipped. If it is, this needs a lot more tests

tests/tests.ts
```diff
@@ -0,0 +1,7 @@
+import './httpLink';
```
jbaxleyiii: Is this file needed? Can you not glob test?

src/link.ts
```diff
@@ -0,0 +1,234 @@
...
+import { DocumentNode } from 'graphql/language/ast';
+
+export abstract class ApolloLink {
+
+  public static from(links: (ApolloLink | RequestHandler)[]) {
+    if (links.length === 0) {
+      return ApolloLink.empty();
+    }
+
+    return links
```
evanshauser: @jbaxleyiii The issue I ran into was that reduce is applied, so that the initial argument is the first in the chain, so the function calls became:

```js
ApolloLink.empty().concat(first).concat(second)... 
```

We could do an ApolloLink.passthrough() as the initial value

src/pollingLink.ts
evanshauser: Sounds good!

src/link.ts
jbaxleyiii: ah shoot, yep! thank you! Lets keep it as is

PR #6, opened 2017-06-15
Initial Apollo Link Specification

evanshauser:
This is the draft specification of the Apollo Link architecture as a standard interface for fetching GraphQL operations. The interface is designed for the current Apollo Client network interface and to provide a simple GraphQL client. Please add your feedback and suggestions on this Pull Request.


README.md
```diff
@@ -1,3 +1,5 @@
 # apollo-fetcher
 
-This is the beginning of apollo-fetcher. The new network interface layer for Apollo Client.
+apollo-fetcher is a standard interface for fetching GraphQL results, designed to support the network interface of Apollo Client and provide a simple standalone GraphQL client capable of extensions.
```
helfer: I don't think we need to mention Apollo Client here at all. We can always add it later.

SPEC.md
```diff
@@ -0,0 +1,225 @@
+## Apollo Fetchers
+
+The purpose of apollo-fetcher is to define an extensible standard interface for fetching GraphQL results. To that end, apollo-fetcher describes an interface containing a single method that connects a GraphQL `Operation` (also called request), to an `Observable` which delivers the results of the operation. Operations that return a single result can easily be mapped to a promise, since the observable's `next` function is called only once. In the general case, Apollo Fetcher uses observables to support GraphQL subscriptions and live queries. The apollo-fetcher's `Observable` follows the ECMAScript [proposal](https://github.com/tc39/proposal-observable), adding a `status` function. A basic fetcher is visualized as follows:
```
helfer: I'd leave out the `status` part for now, and just say that our observable implements the TC39 proposal.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+<p align="center">
+  <br>
+  <img src="images/apollo-fetcher.png" alt="Apollo Fetcher"/>
+</p>
+
+Fetchers are designed for current Apollo Client use-cases and to provide room for extensions. In addition to supporting Apollo Client, Apollo Fetchers are designed to be used independently as a very simple GraphQL client. These are the use cases that fetchers should support:
```
helfer: I'd say that they're designed to meet all the current and anticipated use-cases of Apollo Client, with the possibility of adding extensions in the future.

helfer: I think we're emphasizing Apollo Client way too much in this sentence. I may have said something else previously, but I now think that we should only mention it in passing. Apollo fetchers aren't built for Apollo Client, they're built to be standalone and useful for anyone who needs a simple but extensible GraphQL fetcher.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+* Support middleware to modify headers
+* Support middleware to modify query, variables, operationName and metadata
+* Isomorphic: It works in the browser and on the server
+* Support afterware to modify response
+* Support HTTP  transport, websocket transport, hybrid transport
+* Support alternative transports
+* Support per request retries based on network status or response
+* Support HTTP transport batching
+* Support polling
+* Support batching
```
helfer: I think this is duplicated.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+* Request deduplication of on-the-wire requests
+* Support splitting: eg. send queries and mutations to different servers
+
+Since this list focuses on Apollo Client's needs, please open a Pull Request adding to this list as you think of your desired use cases.
```
helfer: Let's just say "If you can think of more use-cases, please open a PR adding to this list. Make sure to clearly explain your use-case in the PR."

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+In order to support all these use cases with a simple fetcher design, we want to use a modular architecture that allows fetchers to delegate to other fetchers.
+
+## Example
+
+This repository will contain a family of canonical fetchers that provide example extensions. The goal of these canonical fetchers is to be self-contained without changing the context of each operation.
```
helfer: This isn't really pertinent to "Example". Maybe just remove that and the next one.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+The current sample is http-fetcher, which sends an operation over http to a server at a specific URI. The uri and fetch function are passed to the constructor. This Http-fetcher can be used as a stand-alone GraphQL client:
```
helfer: I just realized that if we're passing the `fetch` function to the fetcher, then it doesn't have to be a HTTP fetcher at all, because fetch could do something totally different. Maybe this means that we should clarify that passing a `fetch` function is optional.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+
+```js
+const pollingFetcher = new PollingFetcher(
+  10000, //Polling Interval in ms
+  new HTTPFetcher({ uri: 'http://api.githunt.com/graphql' })
+);
+```
+
+In addition to the query, `Operation` includes a context, which is modified and passed down the stack of fetchers. In addition, this context is sent to the server to through the query body. This context could be used by a polling fetcher to tell a caching fetcher a request should be returned from the cache or not.
```
helfer: I'd say "this context _may_ be sent to the server". It doesn't have to be.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+});
+
+responseObservable.subscribe({
+  next(data){ console.log('received data', data); },
+  error(error){ console.error(error); },
+  complete(){ console.error('request complete'); },
+});
+```
+
+A fetcher can delegate to other fetchers. For example a polling fetcher can send a query operation to an http fetcher on a specified interval. Each request is routed through that http fetcher and returned through the observable returned by the polling fetcher. Using this type of delegation, fetchers can form stacks that each add their own functionality.
```
helfer: Now that I think of it, I'd move this paragraph and the figure below to right before "Example"

SPEC.md
helfer: "In addition to query, **variables and operationName**, ... "

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+The fetcher interface contains a single method:
+
+```js
+interface ApolloFetcher{
+  request(Operation) : Observable<FetchResult>
+}
+```
+
+An Operation contains the AST for a particular query, variables and associated context:
```
helfer: You forgot operationName

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+```js
+FetchResult{
+  data?: object,
+  errors?: object[],
+  extensions?: any,
+  context?:object,
+}
+```
+
+* Should explain what might go into context.
```
helfer: I guess this is a note that was accidentally left over? :wink:

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+* Polling: repeats requests on a specified interval
+
+<p align="center">
+  <br>
+  <img src="images/polling-fetcher.png" alt="Polling Fetcher"/>
+</p>
+
+* Caching: returns data if result in the cache and stores data in cache on response
+* Compose: combines a list of fetchers (would require additional semantics around constructor argument order)
```
helfer: I'm not sure what this one would do. How is it different from split? or is this like the higher-order `compose` function which would let you chain fetchers?

Writing that just made me realize that we're still missing a code example of how you would stack multiple fetchers. I think we should have something like a `compose` function, but we'll have to figure out the details. For now, maybe you can just make a code example of manually connecting them?

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+* Dedup: saves query signatures that are currently on the wire and returns the result for all of those queries
+* NPlusOne: adds query based on predicted access
```
helfer: Why the name NPlusOne? Is it to avoid future queries?

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+Adapters:
+
+* Wrapper => returns a promise : not a fetcher
+
+<p align="center">
+  <br>
+  <img src="images/fetcher-as-promise.png" alt="Fetcher Promise Wrapper"/>
+</p>
+
+## Open questions, discussion
```
helfer: This section is a bit unpolished. Maybe you can clean it up by splitting it into subsections for each discussion topic, and writing a sentence or two about what should be discussed. For example

```
### How to pass context to the `next` callback?
Should we pass the context from the response (for example relevant HTTP headers or status code) to the `next` callback of the observable like this?: `next({ data, errors, extensions, context }){ }` or like this: `{ result: { data, errors, extensions }, context }`.

```

etc.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+> Pass context to the `next` callback along with the query data?
+
+
+Status is complete when finished and if anyone subscribes call complete to avoid memory leaks (other option is error)
+
+subscribe method need to have the Subscriber prototype.
```
helfer: I think people will need a bit more context and details on all of these points here.

README.md
stubailo: I'd maybe go the other way and mention several use cases:

- Fetching queries directly when you don't need normalized caching
- Network interface for Apollo Client
- Network interface for Relay Modern
- Fetcher for GraphiQL

SPEC.md
stubailo: How aligned is the TC39 proposal with RxJS these days?

SPEC.md
stubailo: Agreed. Also I don't know what "context of an operation" means

SPEC.md
stubailo: The current example code doesn't pass `fetch`.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+The current sample is http-fetcher, which sends an operation over http to a server at a specific URI. The uri and fetch function are passed to the constructor. This Http-fetcher can be used as a stand-alone GraphQL client:
+
+<p align="center">
+  <br>
+  <img src="images/http-fetcher.png" alt="Apollo Fetcher"/>
+</p>
+
+```js
+const httpFetcher = new HTTPFetcher({ uri: 'http://api.githunt.com/graphql' });
```
stubailo: I kind of feel like fetchers could be functions instead of classes/instances, and that would be more aligned with the current JS trends.

Why not:

```js
const httpFetcher = createHTTPFetcher({ uri: '...' });
const observable = httpFetcher({ ... });
```

SPEC.md
stubailo: I feel like I prefer the concept of "composing" rather than "delegating". To me, delegating means that a polling fetcher might use an http fetcher internally, but composing means I connect the polling fetcher and http fetcher myself.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+A fetcher can delegate to other fetchers. For example a polling fetcher can send a query operation to an http fetcher on a specified interval. Each request is routed through that http fetcher and returned through the observable returned by the polling fetcher. Using this type of delegation, fetchers can form stacks that each add their own functionality.
+
+<p align="center">
+  <br>
+  <img src="images/polling-stack.png" alt="Http Fetcher"/>
+</p>
+
+
+```js
+const pollingFetcher = new PollingFetcher(
```
stubailo: Similar to the suggestion above, what if we had a more functional approach?

```js
const myFetcher = composeFetchers(
  pollingFetcher(100),
  httpFetcher({ uri: '...' })
);
```

The idea is that it shouldn't be up to the individual fetcher to handle composition, but rather up to the user to decide where they go. But maybe that makes it harder to enforce correct ordering? Not sure.

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+  <br>
+  <img src="images/context.png" alt="Fetcher Stack"/>
+</p>
+
+## API
+
+The fetcher interface contains a single method:
+
+```js
+interface ApolloFetcher{
```
stubailo: Do we want to stick to TypeScript syntax here? This isn't currently valid, not that it matters that much.

SPEC.md
stubailo: I think this sentence is redundant with the interface below

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+
+```js
+Observable{
+  //Causes the underlying producer to start
+  subscribe(Subscriber<FetchResult>) : UnsubscribeHandler
+  subscribe(
+    next(FetcheResult) => void,
+    error(any) => void,
+    complete() => void) : UnsubscribeHandler
+  status(): any
```
stubailo: What's the deal with this status thing?

SPEC.md
```diff
@@ -0,0 +1,225 @@
...
+  next(T),
+  error(any),
+  complete(),
+}
+```
+
+UnsubscribeHandler is returned from a `subscribe` function in the Observable
+
+```js
+type UnscubscribeHandler = () => void
```
stubailo: Typo in name

SPEC.md
stubailo: Totally agree

SPEC.md
jbaxleyiii: I'd say we could go even further. 

```js
const network: Fetcher = HTTPFetcher({ uri: '...' });
const polling: Fetcher = PollingFetcher({ interval: 100 });

const fetcher = network.concat(polling);
// or
const fetcher = Fetcher.from([ network, polling ]);
// or
const fetcher = Fetcher.of(network).concat(polling);
```

If possible, does it make sense to stick to algebraic structures? Or is that too confusing for people? If you did, you can do things like this:

```js
// create the fetcher
const fetcher: Fetcher = HTTPFetcher({ uri:  'http://api.githunt.com/graphql' })
   // don't send requests with client directive
  .filter(request => !isClientOperation(request.operation))
  // add authorization (middleware)
  .map(request => {
    request.context.headers.authorization = userId();
    return request;
  });

// operationName can deduced from the operation
// this part won't normally be exposed to the end user but used in apollo-client or other libs
// should this be positional vs destructured
const observable = fetcher.request({ operation: DocumentNode, variables: Object, context: Object });

const subscription  = observable
  // don't notify if offline error
  .filter((data, error) => error instanceof NetworkError)
  // change shape of data (afterware)
  .map(data => data.items.map(x => localize(x.date)))
  .subscribe(
    // data
    data => console.log('received data', data),
    // error
    error => console.error(error),
    // complete
    () => console.error('request complete'),
  );

// When we're done
subscription.dispose();

```

From what I can tell from the (excellent!) diagrams, the critical bits are the request information and the result information. Each fetcher should be able to modify this information (forming middlewares and afterwares), and should be able to build on top of each other.

The `Fetcher` interface should allow for control flow based on the `Request` information (i.e operation, variables, context). One should be able to map, filter, and concat the Fetcher to modify request information before it is sent. Since the fetcher is returned from each function, a specific fetcher would be able to add things to the interface like `interval` or `batchIf` or etc. I think there is a lot that can be learned from RxJS in building enjoyable observable interfaces.

Upon `.request`

The `Observable` should allow for control flow related to the result of the request (i.e. data and error). One should be able to map, filter, and concat based on result and error data. This forms the basis for afterware and allows fine grained control of subscription handler firings.

In fact, I wonder if looking at something like [RxJS.lite](https://www.npmjs.com/package/rx-lite) might even be a good starting point for the generic Fetcher library. Things like `catch`, `interval`, and `retry` seem like perfect fits for fetchers.

I love the idea of fetchers being used on their own as a slim build of a graphql library. Apollo Client itself could be rebuilt to be essentially the ~fully built~ combination of excellent fetchers. Having more of a functional approach to the way Fetchers are defined and used could make this natural and allow for interop between compile to languages (like reason) and give strong type inference for things like typescript and flow.

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+interface ApolloFetcher{
+  request(Operation) : Observable<FetchResult>
+}
+```
+
+An `Operation` contains all the information necessary to execute a GraphQL query, including the GraphQL AST:
+
+```js
+Operation {
+  query: DocumentNode,
```
jbaxleyiii: I don't think query should the key. For things like mutations and subscriptions, this seems off. Why not `operation` and call the type `Request`

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+  operationName: string,
```
jbaxleyiii: since we have the AST, we can deduce operationName

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+```js
+Observable{
+  //Causes the underlying producer to start
+  subscribe(Subscriber<FetchResult>) : UnsubscribeHandler
+  subscribe(
+    next(FetcheResult) => void,
+    error(any) => void,
+    complete() => void) : UnsubscribeHandler
+}
```
jbaxleyiii: just a heads up this has some syntax errors

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+```js
+Subscriber<T>{
+  next(T),
+  error(any),
+  complete(),
+}
+```
+
+`UnsubscribeHandler` is returned from `subscribe`
```
jbaxleyiii: To keep with spec, shouldn't this return a `Subscription` which can be unsubscribed and determined if closed? https://github.com/tc39/proposal-observable#api

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+```js
+type UnsubscribeHandler = () => void
+```
+
+`FetchResult` is passed to the `next` callback of `Subscriber`
```
jbaxleyiii: Why would you pass all of this to `next`? Shouldn't `next` get just data, `error` get the error? What are the use cases around extentions and context when you have a result?

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+* Caching: returns data if result in the cache and stores data in cache on response
+* Compose: combines a list of fetchers (would require additional semantics around constructor argument order)
+* Dedup: saves query signatures that are currently on the wire and returns the result for all of those queries
+* Retry: error callback causes the request to retry
+* Mock: returns fake data for all or part of a request ← implement with BaseFetcher (eg. executes `graphql`)
+
+Fork:
+
+* Split: split operations between fetchers depending on a function passed in
```
jbaxleyiii: I think writing out how the API for this could look would be really helpful in the overall design. Splitting is an interesting concept and may be possible simply by allowing fetchers to filter which requests they want to handle. 

```js
const XHR = HTTPFetcher({ uri: '...' })
  .filter(({ operation })=> isNotSubscription(operation))
const WS = WSFetcher({ socket: '...' })
  .filter(({ operation }) => isSubscription(operation))

const fetcher = Fetcher.from([XHR, WS]);
```

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+  completed
+}
+```
+
+If `status` exists and an `Observable` has terminated, should `subscribe` throw an error or call complete immediately, avoiding memory leaks.
+
+### Does the subscribe method need to be overloaded?
+
+Currently the `subscribe` method has two different signatures, one that takes three functions `next`, `error`, and `complete` and antoher that takes a `Subscriber`.
+Positional arguments are more prone to errors than passing object containing arguments.
```
jbaxleyiii: I'm not sure I agree with this. I prefer positional arguments when the needs of the arguments are hierarchical. That being said, the spec calls for using an object so I'd go with that and not the overload https://github.com/tc39/proposal-observable#api

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+* Polling: repeats requests on a specified interval
+
+<p align="center">
+  <br>
+  <img src="images/polling-fetcher.png" alt="Polling Fetcher"/>
+</p>
+
+* Caching: returns data if result in the cache and stores data in cache on response
+* Compose: combines a list of fetchers (would require additional semantics around constructor argument order)
+* Dedup: saves query signatures that are currently on the wire and returns the result for all of those queries
```
jbaxleyiii: Things like `Dedup`, `Retry`, and one I'd like to see (`Catch`) don't make sense as fetchers to me, but rather middle / afterwares which run the fetcher call stack in a modified way. They aren't transport specific (i.e where the data comes from) but are operational (i.e. how to call the request). I'd rather see a `Fetcher` as a way to get data and it have the ability for its control flow to be modified by other utilities. This would make things like logging and analytics easy to integrate as well since they are essentially side effects.

SPEC.md
```diff
@@ -0,0 +1,258 @@
...
+
+http-fetcher sends an operation over HTTP to a server at a constructor specified URI. Additionally, the constructor accepts an optional fetch function and defaults to isomorphic-fetch. A custom fetch function would contain the HTTP specific behavior normally found in middleware or afterware. In this manner, http-fetcher can function as a stand-alone GraphQL client:
+
+<p align="center">
+  <br>
+  <img src="images/http-fetcher.png" alt="Http Fetcher"/>
+</p>
+
+```js
+const httpFetcher = new HTTPFetcher({ uri: 'http://api.githunt.com/graphql' });
```
jbaxleyiii: This is a personal preference, but I prefer to avoid using `new` where possible. It can be confusing to beginners to know when to use it and aside from `classes`, is not needed for standard function calls.

SPEC.md
stubailo: How would one implement polling in this model? If the polling fetcher/middleware needs to be able to call into the HTTP one?

SPEC.md
stubailo: Well, in most HTTP implementations it's called `query` on the wire. I think that's a pretty big oversight in GraphQL but we're kinda stuck with it now.

SPEC.md
stubailo: Nof if there are multiple operations in the document

SPEC.md
stubailo: Note: In this design both middlewares and afterwares are just called "Fetchers"

jbaxleyiii:
@evanshauser I wonder if the goal of making everything a fetcher is confusing. I think with a network interface, you are dealing with two things. The transport layer, and the control flow around it.

A transport layer is responsible for going and getting data (i.e. HTTP request, WS message, or even service worker / cache lookup). The control flow is about how to execution the operation (i.e. polling, retry, catching on error, batching).

Composition in this case is about choosing when a query gets sent to a transport, and how to execute the operation (and when). Being able to filter on a transport allows for selective fetching (if authenticated, WS vs HTTP vs cache). Being able to map on it allows changing the request itself (adding things to the header for instance).

The observable represents the future fulfillment of data so it's a great place to modify the results of the transport(s) to prep the data for usage by the application / storage in cache.

Reading over this I was confused about what was an `extension` and what was an actual usable fetcher on its own. I can't use a `pollingfetcher` without an httpfetcher right? They seem like different things to me.

SPEC.md
jbaxleyiii: That does stink

SPEC.md
jbaxleyiii: if there are multiple operations, wouldn't you want multiple operationNames? The AST would list all operations. What is the uses case of operationName here?

SPEC.md
jbaxleyiii: I added a note at the bottom. Since middlewares / afterwares deal with modification of information before / after a request is made, it seems odd that a `Fetcher` would be both middle / after ware, and the method of making the request. I may be getting hung up on the naming here but I guess I see them as different things to be done

SPEC.md
stubailo:
```
query FirstQuery {
  asdf
}

query SecondQuery {
  jkl
}
```

When you send this operation to the server, you need to include an `operationName`.

SPEC.md
stubailo: Yeah, I think the naming is a bit confusing in that way.

SPEC.md
jbaxleyiii:
```js
const network: Fetcher = HTTPFetcher({ uri: '...' });

function PollingFetcher({ interval }){
  // simplified version of what would need to be done

  // this is the HTTP request
  const originalRequest = this.request;
  this.request = () => {
    setInterval(originalRequest, interval);
  }

  return this
}

const polling: Fetcher = PollingFetcher({ interval: 100 });

const fetcher = network.concat(polling);
```

SPEC.md
jbaxleyiii: I didn't realize that you could send multiple operations to a server? I thought to batch them they needed to be a single operation or an array of operations?

jbaxleyiii: For instance https://launchpad.graphql.com/z0915lrp7 only allows one operation at a time

SPEC.md
stubailo:
```js
const polling = (interval) => ({ next, onUnsubscribe }) => {
  const interval = setInterval(next, interval);
  onUsubscribe(() => clearInterval(interval));
}

const network = createHttpFetcher({ uri: '...' });

const fetcher = compose(polling(100), network);
```

stubailo: I might be getting too excited about functional junk here.

jbaxleyiii: The hard part is if an Fetcher needs to implement `request` what does the request of `PollingFetcher` look like on its own? It doesn't actually have anything to request

stubailo: Sorry I'm still on my syntax above. My functional composition ideas only work if those things are actually functions.

jbaxleyiii: Haha comments got out of order here. We commented at the same time so it looks like I replied to you when I meant to reply to myself

helfer:
@jbaxleyiii @evanshauser @stubailo There's some great feedback here. Let's talk about this tomorrow during contributor's club!

evanshauser:
@jbaxleyiii This is awesome! Thank you for your thoughts. We definitely can use some of these ideas. I really like the thought of adding more functionality to the Observable. By providing `map`, `filter`, `catch`, and other functions, implementing a custom fetcher would become much easier. I'll have to think about how the additions incorporate. My first though is as optional functions on Observable and concrete implementations in an AbstractObservable.

Breaking the composition up into `Fetcher.of().concat()` makes sense for simple fetcher stacks. Splitting gets a bit more complicated, something like this might work:

```js
const mutationFetcher = Fetcher.of(...).concat(...);
const queryFetcher = Fetcher.of(...).concat(...)

Fetcher.of(...).concat(...).split((operation) => {
  if(FetcherUtils.hasMutation(operation)){
    return mutationFetcher;
  } else {
    return queryFetcher;
  }
)

//or

Fetcher.of(...).concat(...).split((operation, fetchers) => {
  if(FetcherUtils.hasMutation(operation)){
    return fetcher[0];
  } else {
    return fetcher[1];
  },
  Fetcher.of(...).concat(...),
  Fetcher.of(...).concat(...),
)
```

To support `Fetcher` we would need to impose some ordering on the constructor arguments. I'll leave room to create the `.of()` functionality when I create the BaseFetcher examples. Let me know your thoughts!

> `Fetcher` could be named `AFetcher` or `ApolloFetcher`

SPEC.md
```diff
@@ -0,0 +1,322 @@
...
+Adapter (not a Fetcher):
+
+* Promise Wrapper
+
+<p align="center">
+  <br>
+  <img src="images/fetcher-as-promise.png" alt="Fetcher Promise Wrapper"/>
+</p>
+
+* Backwards-compatablility Wrapper: exposes `query`, `mutate`, and `subscribe`
```
jnwng: sp

SPEC.md
```diff
@@ -0,0 +1,426 @@
...
+A basic link is visualized as follows:
+
+<p align="center">
+  <br>
+  <img src="images/apollo-link.png" alt="Apollo Link"/>
+</p>
+
+### Modular Architecture
+
+To support a range of use cases with simple links, Apollo Links are designed as a modular system that composes links, chaining together their functionality.
```
evanshauser: that composes links together forming chains of functionality

SPEC.md
```diff
@@ -0,0 +1,426 @@
...
+* Do a GraphQL query that contains @defer, @stream or @live
+  * Support custom annotations, such as @connection
+
+<br>
+
+* Modify headers as middleware currently does
+  * Add authentication headers
+* Modify query, variables, operationName and context
+  * Use persistent queries
+* Support afterware to modify response or add side effects
```
evanshauser: Afterware and Middleware are now responsibilities of apollo-fetch

docs/roadmap.md
```diff
@@ -0,0 +1,177 @@
...
+  * transform network error into GraphQL error
+
+<br>
+
+* Network layer caching
+  * Denormalized
+  * Normalized
+  * Local Storage
+  * IndexDB
+
```
jbaxleyiii: I think we should include a use case here about local client state as a teaser of what is to come 😉

docs/roadmap.md
```diff
@@ -0,0 +1,177 @@
...
+If you think of more use-cases, please open a PR adding to this list and make sure to clearly explain your use-case.
+
+## Open questions, discussion
+
+Throughout the design process, a couple questions have surfaced that may prompt conversation.
+If you want to see something here, please open a PR with your proposed change or an Issue for discussion.
+
+<!--
+### How should we pass context to `next` callback along with the query data?
```
jbaxleyiii: I'd just remove this since it isn't needed anymore

docs/roadmap.md
```diff
@@ -0,0 +1,177 @@
...
+  started
+  stopped
+  errored
+  completed
+}
+```
+
+If `status` exists and an `Observable` has terminated, should `subscribe` throw an error or call complete immediately, avoiding memory leaks.
+
+<!--
```
jbaxleyiii: same for this and all other commented out portions

docs/roadmap.md
```diff
@@ -0,0 +1,177 @@
...
+### Which convenience functions would you like?
+
+The current functions provided:
+
+* ApolloLink
+  * asPromiseWrapper
+
+Proposed additions include:
+
+* Pull information from the Operation as part of `Operation` or `LinkUtil` library
```
jbaxleyiii: I think these would be incredible as AST helpers

docs/roadmap.md
```diff
@@ -0,0 +1,177 @@
...
+
+<!--
+* Observable additions as required or optional part of interface and implemented in `AbstractObservable`
+  * `map`
+  * `filter`
+  * `catch`
+  * `finally`
+-->
+
+### Should GraphQL errors be propagated up the stack with `next` or `error`?
```
jbaxleyiii: I think this distinction is great since a graphql error can also return data. Previously we couldn't handle this situation, now we can

docs/summary.md
```diff
@@ -0,0 +1,207 @@
...
+```js
+import {
+  execute,
+  HttpLink,
+} from 'apollo-link';
+
+const uri = 'http://api.githunt.com/graphql';
+const link = new HttpLink({ uri });
+
+execute(link, operation).subscribe({
```
jbaxleyiii: this makes my heart so happy

docs/summary.md
```diff
@@ -0,0 +1,207 @@
...
+
+const uri = 'http://api.githunt.com/graphql';
+const context = {
+  headers: {
+    auth: 'token',
+  },
+};
+
+const link = ApolloLink.from([
+  new SetContextLink(context),
```
jbaxleyiii: hmm, this makes me wonder if setContext should be a function? so it can be run for each request?

docs/summary.md
```diff
@@ -0,0 +1,207 @@
...
+
+### SetContextLink
+
+```js
+new SetContextLink( context );
+```
+
+* `context` sets the context of the operation, which other links can access
+
+### MockLink
```
jbaxleyiii: is this something we are shipping?

docs/summary.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+* `max` is the maximum number of requests for an operation, defaults to 10
+* `delay` is used to calculate the time between retries, defaults to 300
+* `interval(delay, count)` calculates the time defaults to returning `delay`
+
+### SetContextLink
+
+```js
+new SetContextLink( context );
```
jbaxleyiii: see earlier comment about this being a function so it can be executed each time

docs/summary.md
evanshauser: This is great!

docs/roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+* Polling
+* Batching at the transport/fetch layer
+  * Perform batch operations within a composed link
+* Deduplicate of on-the-wire requests
+* Re-route and modify errors
+  * transform network error into GraphQL error
+
+<br>
+
+* Network layer caching
```
nevir: Feedback from the outside:

I find this point pretty interesting - are you guys thinking in terms of being able to effectively invert the relationship between apollo-link and apollo-client in the future?  E.g. apollo-client's normalization logic becomes a link - and allowing consumers to effectively pick and choose the behaviors they want their client to have?

docs/roadmap.md
```diff
@@ -0,0 +1,122 @@
...
+* Pull information from the Operation as part of `Operation` or `LinkUtil` library
+  * hasQuery
+  * hasMutation
+  * hasSubscription
+  * getQuery
+  * getMutation
+  * getSubscription
+  * annotation support?
+
+### Should Link include some form of synchronous adapter/behavior? What use cases warrant this?
```
nevir: I think you're going to find a strong need for synchronous behavior at some level to maintain compatibility with projects that rely on highly synchronous stores such as Redux.  It may only need to be synchronous at the caching layer (e.g. `readQuery`/`writeQuery` from apollo-client) though; not from apollo-link's perspective

docs/roadmap.md
evanshauser: Yeah! Thank you for adding to the discussion. Allowing consumers to pick functionality is certainly a goal. Inverting the relationship is a possible future option, meaning apollo-client could become the synchronous adapter on top of a bunch of `apollo-link`s. Any further thoughts would be really appreciated and informative for future development

docs/implementation.md
```diff
@@ -0,0 +1,210 @@
+## Implementing Links
+
+Before implementing your own Link, be sure to read the [usage](usage.md) document.
+
+Links implement a `request` method that takes an operation and returns the results in an Observable.
+The apollo-link's `Observable` follows the ECMAScript [proposal](https://github.com/tc39/proposal-observable#api) with some extensions provided by [zen-observable](https://github.com/zenparsing/zen-observable#api).
+
+### Creating a Simple Http Link
+
+All links extend the base class, `ApolloLink`, which contains the methods for composition as well as the signature of request.
```
nevir: None of the composition methods inherited from `ApolloLink` seem particularly coupled to the implementation of each specific Link class.  I wonder if it might be cleaner to Make `ApolloLink` a pure interface, and move the composition helpers out to a separate object.  You're already moving in that direction with static helpers like `ApolloLink.from`

docs/implementation.md
```diff
@@ -0,0 +1,210 @@
+## Implementing Links
+
+Before implementing your own Link, be sure to read the [usage](usage.md) document.
+
+Links implement a `request` method that takes an operation and returns the results in an Observable.
+The apollo-link's `Observable` follows the ECMAScript [proposal](https://github.com/tc39/proposal-observable#api) with some extensions provided by [zen-observable](https://github.com/zenparsing/zen-observable#api).
```
nevir: I wonder if it's worth a _further_ extension of exposing a `then` and `catch` for `Promise`-like semantics for the simple cases where a caller just wants to make a straight network request w/o any caching or subscriptions

docs/implementation.md
```diff
@@ -0,0 +1,210 @@
...
+```js
+class HttpLink extends ApolloLink {
+
+  constructor(uri){
+    super();
+    this.apolloFetch = createApolloFetch({ uri });
+  }
+
+  request(operation) {
+    return new Observable(observer => {
```
nevir: One thing to consider is that links may feel the need/want to expose some additional behavior, such as `refetch()`.  Or maybe that's out of scope of a library like this, and where something like apollo-client comes in

docs/implementation.md
nevir: Edit: Ah! https://github.com/apollographql/apollo-link/pull/6/files#diff-438fd70aa65accf70711cf42abd07386R105

README.md
```diff
@@ -1,3 +1,31 @@
 # apollo-link
 
-This is the beginning of apollo-link. The new network interface layer for Apollo Client.
+`apollo-link` is a standard interface for fetching GraphQL results, designed to provide a simple GraphQL client that is capable of extensions.
+High level use cases are:
```
stubailo: This seems somewhat unfinished. I'd use more complete sentences, like "Here are the use cases where we think Apollo Link will be a good fit:"

README.md
```diff
@@ -1,3 +1,31 @@
...
+* network interface for Relay Modern
+* fetcher for GraphiQL
+
+## Documentation
+
+To start, begin by reading the getting started [guide](docs/summary.md).
+
+If you are interested in implementing your own links, read the implementation [information](docs/implementation.md)
+
+The current plans and furture implementation are in the [roadmap](docs/roadmap.md)
```
stubailo: furture -> future

README.md
```diff
@@ -1,3 +1,31 @@
...
+
+Your feedback and contribuitions are welcome.
```
stubailo: contributions

README.md
```diff
@@ -1,3 +1,31 @@
...
+
+## Apollo Principles
+
+`apollo-link` strives to follow the Apollo design principles:
+
+1. **Incrementally adoptable**
```
stubailo: I don't think these need to be bold

docs/summary.md
```diff
@@ -0,0 +1,203 @@
+## Apollo Link
+
+Apollo Link is an extensible standard interface for modifying control flow of GraphQL queries and fetching GraphQL results.
+
+## Basic Usage
+
+Links can be used as a stand-alone client or with most major GraphQL clients.
+
+### Stand-alone
```
stubailo: I'd put a code sample above any actual documentation. So maybe above the `standalone` header.

docs/summary.md
```diff
@@ -0,0 +1,203 @@
...
+
+`execute` accepts a standard GraphQL request with a query string or AST.
+`execute` returns an [Observable](https://github.com/zenparsing/zen-observable#api) that can be subscribed to.
+Links use observables to support GraphQL subscriptions and live queries, in addition to single response queries and mutations.
```
stubailo: Not just for subscriptions and live queries - also for polling! (meaning it's still valuable if you're just using HTTP)

docs/summary.md
stubailo: Can we add an example of using this as a promise as well?

docs/summary.md
```diff
@@ -0,0 +1,203 @@
...
+
+If using `create-react-app`, your `index.js` would look similar to above.
+
+## Adding Functionality
+
+Links can composed together to form a new link using `ApolloLink.from`.
+
+### Retry
+
+Attempts to resend a GraphQL request when failed and fails after a certain time
```
stubailo: Complete sentence here.

docs/summary.md
```diff
@@ -0,0 +1,203 @@
...
+<GraphiQL fetcher={(operation) => execute(link, operation)}/>,
+
+const client = new ApolloClient({
+  networkInterface: link,
+});
+```
+
+### Authentication
+
+Adds authentication headers, using the context
```
stubailo: I think this is missing a bit of an explanation that `HttpLink` ends up reading the context to get the headers.

docs/summary.md
```diff
@@ -0,0 +1,203 @@
...
+
+const link = ApolloLink.from([
+  new SetContextLink(setContext),
+  new HttpLink({ uri });
+])
+
+<GraphiQL fetcher={(operation) => execute(link, operation)}/>,
+```
+
+### Retry Authenticated Operation
```
stubailo: Maybe "Example: Combining links"?

docs/summary.md
```diff
@@ -0,0 +1,203 @@
...
+  operation: {
+  query: DocumentNode | string,
+  operationName?: string
+  variables?: object
+  context?: object
+  },
+)
+```
+
+### HttpLink
```
stubailo: If this accepts options from the context, we should document that

docs/summary.md
```diff
@@ -0,0 +1,203 @@
...
+
+### RetryLink
+
+```js
+new RetryLink({ max, delay, interval });
+```
+
+* `max` is the maximum number of requests for an operation, defaults to 10
+* `delay` is used to calculate the time between retries, defaults to 300
+* `interval(delay, count)` calculates the time defaults to returning `delay`
```
stubailo: I'm not sure what that means

docs/implementation.md
stubailo: Ooh, that's interesting. Maybe a future version should have ways for that to be exposed to the user.

docs/implementation.md
```diff
@@ -0,0 +1,210 @@
...
+Otherwise, the composing function will warn the user.
+This distinction and check is designed to ensure that users who import a Link will not be surprised if an operation is not forwarded.
+
+```js
+request() //terminating
+request(operation) //terminating
+request(operation, forward) //non-terminating
+```
+
+### Single Path - from and concat
```
stubailo: Maybe "combining links"?

docs/implementation.md
```diff
@@ -0,0 +1,210 @@
...
+</p>
+
+The default `right` argument is a passthrough link that will call `forward(operation)` and return an empty Observable otherwise.
+A `split` can become a filter by setting `right` to `ApolloLink.empty()`.
+
+```js
+ApolloLink.split(
+  test: (Operation) => boolean,
+  left,
+  right? = ApolloLink.passthrough(),
```
stubailo: Do we actually want people to use this? I feel like it's more confusion than just putting the chain here.

docs/roadmap.md
stubailo: > allowing consumers to effectively pick and choose the behaviors they want their client to have?

Definitely!

docs/implementation.md
evanshauser: The first thought was to keep `asPromiseWrapper` out of apollo-link, since simple queries can be done with [apollo-fetch](https://github.com/apollographql/apollo-fetch). You make a good point that it's worth adding `asPromiseWrapper` back in. Any reasons not to @jbaxleyiii

docs/implementation.md
evanshauser: Good thought! It would make the `ApolloLink` into an interface. The reason to included `split` and `concat` on `ApolloLink` is to make long compositions in user code more readable/clean. Also users won't have to import ApolloLink if Links are used across different files.

I think this composition is more understandable than nesting `from`s and the `split`

```js
const link = loggingLink
.split(
  isLocal,
  checkLocalCache,
  ApolloLink.passthrough(),
)
.concat(dedup)
.concat(retry)
.concat(http)
```

stubailo:
```js
const link = ApolloLink.from([
  loggingLink,
  ApolloLink.split(
    isLocal,
    checkLocalCache,
    ApolloLink.passthrough(),
  ),
  dedup,
  retry,
  http,
]);
```

I feel like it's fine either way.

docs/summary.md
```diff
@@ -26,8 +20,16 @@ execute(link, operation).subscribe({
   error: error => console.log(`received error ${error}`),
   complete: () => console.log(`complete`),
 })
+
+//For single execution operations, a Promise can be used
+new Promise((resolve, reject) => (execute(link, operation)).subscribe(resolve, reject))
```
stubailo: Hmm, is there no way to get a promise in a first-class way?

docs/summary.md
```diff
@@ -191,7 +204,7 @@ new RetryLink({ max, delay, interval });
 
 * `max` is the maximum number of requests for an operation, defaults to 10
 * `delay` is used to calculate the time between retries, defaults to 300
-* `interval(delay, count)` calculates the time defaults to returning `delay`
+* `interval(delay, count) => number` calculates the time between retries, defaults to returning `delay`
```
stubailo: Is `count` the number of retries that have already happened?

docs/summary.md
jbaxleyiii: @stubailo @evanshauser's had the idea to export a `toPromise(execute(link, operation))` method. Would that be enough?

We could also do `executeToPromise` but I don't like that nearly as much

jbaxleyiii:
@evanshauser aside from @stubailo's remaining comments, I think this is wonderful and we should merge it and release today! 🎉

docs/summary.md
stubailo: I just think most people are going to want a promise in the standalone use case. But we can add it later!

stubailo:
Yeah, it's awesome! I think everything else can be improved later as we go.

docs/implementation.md
nevir: You could take the builder approach, if chaining is preferable.  Something like:

```ts
const link = ApolloLink
  .chain(loggingLink)
  .split(
    isLocal,
    checkLocalCache,
    ApolloLink.passthrough(),
  )
  .concat(dedup)
  .concat(retry)
  .concat(http)
  .link();
```

docs/implementation.md
jbaxleyiii: @evanshauser as long as there is a clear use case, it is such little code we should make it easier for people to use! (we did that with `compose` and react-Apollo)

PR #7, opened 2017-07-07
NOT FOR MERGE

jbaxleyiii:
# DO NOT MERGE

@evanshauser I took some time to write out a different approach to the link composition and interface as a way to think through the use cases and to better understand the challenges of the link API design. Let's talk through the differences between this implementation and yours next week when I'm in SF. They are both valid approaches but very different in a lot of ways. I think it could be valuable / fun to talk through the pitfalls / strengths of both approaches.

In particular, take a look at the tests and the implementation of `skip`, the `HttpLink`, and the `RetryLink`

Also, I'm not sure why travis is failing, these test are all working great for me locally? (I'm running node 8 which may be part of it?)


src/index.ts
```diff
@@ -1,7 +1,94 @@
...
+    return new Link(operation => this.request(f(operation)));
+  }
+
+  concat(f: Link) {
+    return new Link(operation => {
+      return f.request(operation, this);
+    });
+  }
+
+  filter(test: (x: OperationRequest) => boolean) {
```
stubailo: I think these are duplicative with `Link.filter` - I prefer the static version only.

src/index.ts
```diff
@@ -1,7 +1,94 @@
...
+        context: result.context ? result.context : operation.context || {},
+      };
+    });
+  }
+
+  map(f: (x: OperationRequest) => OperationRequest) {
+    return new Link(operation => this.request(f(operation)));
+  }
+
+  concat(f: Link) {
```
stubailo: I feel we should pick one approach - either Link.from or concat, but not both; at least to start.

src/links/HttpLink.ts
```diff
@@ -0,0 +1,35 @@
...
+import Link from '../';
+
+export default (uri: string) => {
+  // create HTTP utility
+  const fetcher = createApolloFetch({ uri });
+
+  // create Link from request and previous link
+  return new Link((request, prev = Link.empty()) => {
+    // hook up Observerable
+    return new Observerable(observer => {
```
evanshauser: Really like this way of constructing Observables, I think we should use it in our own implementation of Observable or just use zen-observable

src/links/Retry.ts
```diff
@@ -0,0 +1,64 @@
...
+
+// linear interval
+export const defaultInterval = (delay, count) => delay;
+
+// example using a class instead of a closure
+export default class Retry extends Link {
+  max: number = 10;
+  delay: number = 300;
+  count: number = 0;
+  interval: (delay: number, count: number) => number;
```
evanshauser: Awesome interface for the retryLink. Let's use this in our standard Link library

tsconfig.json
```diff
@@ -15,7 +15,6 @@
     "skipLibCheck": true
   },
   "files": [
-    "src/index.ts",
-    "tests/index.ts"
+    "src/index.ts"
```
evanshauser: not including the test files in the compilation may be why the tests were failing. I'll have to check locally first

src/links/HttpLink.ts
stubailo: As long as you call it `Observable` and not `Observerable` 😉

jbaxleyiii: @stubailo haha well it is `going` to a server ;)

Fixed 👍

src/links/Retry.ts
jbaxleyiii: @evanshauser I pushed some changes to it that should make it a little bit more powerful. I'm loving the potential of this project 😍

jbaxleyiii:
@stubailo @evanshauser on the plane last night I tried to go through and document / explain all of the choices in this PR. Looking forward to going to over it in person today 🎉

src/index.ts
jbaxleyiii: @stubailo I feel like they serve different purposes, but I'd love to talk through it!

src/index.ts
jbaxleyiii: Which static version?

src/links/HttpLink.ts
jbaxleyiii: @evanshauser yeah I don't feel like we need to build an observable implementation? It feels like unnecessary work that we would have to maintain instead of relying on a project solely focused on it.

I like the size and narrow scope of Zen vs RxJSLite

Issue #16, opened 2017-07-26
Client that supports subscription using Link


KadoBOT:
Besides the other things that this package can do, it can replace `subscriptions-transport-ws` with the `execute` right? If so, can you guys share an example how create a `client` that supports subscription for `ApolloProvider` using `apollo-link`? I've tried to follow the docs, replacing the `networkInterface` with the `new HttpLink({uri})` and I'm getting `_this.networkInterface.subscribe is not a function`.

Thanks,

evanshauser:
@KadoBOT Thanks for opening an issue! Subscription support is next on our roadmap. We'll nee a WebsocketLink that uses `subscriptions-transport-ws` and a compatibility layer in Apollo Client. Are you interested in contributing to either effort?

KadoBOT:
Sure, why not. :)

evanshauser:
@KadoBOT Great! The initial implementation of the websocket link is on the apollo-link-ws branch. It needs some testing if you are interested or know of any good websocket mocking tools. Your feedback on the code would also be great!

Or if you are interested in the Apollo Client side of things, we have to make a change [here](https://github.com/apollographql/apollo-client/blob/master/src/ApolloClient.ts#L231) to include the subscribe method on the networkInterface adapter.

KadoBOT:
Btw @evanshauser do you think you can help me with an issue that I've opened on apollo-graphlq a few days ago?
Thanks.

stubailo:
@KadoBOT can you link to the issue please?

KadoBOT:
Hi @stubailo yes I can. And it was on apollo-client: https://github.com/apollographql/apollo-client/issues/1957
I'm trying to solve this for two weeks already.

evanshauser:
@KadoBOT WebSocketLink enables subscription support in a Link composition. Will add documentation soon.

PR #30, opened 2017-07-31
Websocket Link with Subscription Support

evanshauser:
This contains the websocket implementation of a link, which depends on [`subscriptions-transport-ws`](https://github.com/apollographql/subscriptions-transport-ws). Currently the implementation needs testing.

Additionally, this PR adds a field in Operation for the type of the operation that will be performed. This needs to be documented. Currently the GraphQL spec says that an AST that describes multiple operations must also include an operationName to indicate which operation to perform.

TODO:


packages/apollo-link-ws/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "postbuild": "cp -R ./dist/src/. ./lib",
+    "watch": "tsc -w -p .",
+    "clean": "npm run clean:dist && npm run clean:coverage",
+    "clean:dist": "rimraf dist/* && rimraf lib/*",
+    "clean:coverage": "rimraf coverage/*",
+    "prepublishOnly": "npm run clean && npm run build"
+  },
+  "dependencies": {
+    "apollo-link-core": "^0.1.1",
+    "graphql": "^0.10.5",
```
DxCx: should be peerDependancy..
anyway, you don't have to use print as far as i know

evanshauser: Sweet! We can just use `subscriptions-transport-ws` directly as a link, since it has the same signature. This is incredible!

CC @stubailo

packages/apollo-link-core/tests/link.ts
```diff
@@ -794,7 +794,7 @@ describe('Link static library', () => {
           assert.notProperty(operation, 'variables');
           assert.notProperty(operation, 'context');
           assert.property(op, 'query');
-          assert.property(op, 'operationName');
+          // assert.property(op, 'operationName'); isn't automatically set if not present
```
stubailo: What's this about?

packages/apollo-link-ws/src/webSocketLink.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { SubscriptionClient, ClientOptions } from 'subscriptions-transport-ws';
+
+export type WebSocketParams = {
+  uri: string;
+  options?: ClientOptions;
+  webSocketImpl?: any;
+};
+
+/** Transforms Operation for into HTTP results.
```
stubailo: This doesn't seem to be the right comment since it refers to HTTP?

stubailo:
Looks great, some odd comments though.

evanshauser:
This fixes #16

evanshauser:
@dozoisch I remember we talked at a React Wednesday(couple weeks ago at MDG) about some issues with subscriptions and reconnecting on a new user session. I wanted to make sure we addressed your use case with the WebsocketLink.

Happy to DM on the [Apollo Slack](http://www.apollodata.com/#slack) as well.

PR #33, opened 2017-08-04
apollo-link-dedup

evanshauser:
This adds a deduplication link that keeps track of operations that are passed to the next link and holds identical operations.


packages/apollo-link-dedup/src/dedupLink.ts
```diff
@@ -0,0 +1,57 @@
...
+  Observable,
+} from 'apollo-link-core';
+
+import { print } from 'graphql/language/printer';
+
+/*
+ * Expects context to contain the deduplicate field
+ */
+export class DedupLink extends ApolloLink {
+  private inFlightRequestPromises: { [key: string]: Observable<FetchResult> };
```
stubailo: This is probably misnamed right - since it's not a promise?

packages/apollo-link-dedup/src/dedupLink.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor() {
+    super();
+    this.inFlightRequestPromises = {};
+  }
+
+  public request(
+    operation: Operation,
+    forward: NextLink,
+  ): Observable<FetchResult> {
+    // sometimes we might not want to deduplicate a request, for example when we want to force fetch it.
```
stubailo: Should it be opt-in, or opt-out?

packages/apollo-link-dedup/tests/dedupLink.ts
```diff
@@ -0,0 +1,14 @@
...
+import * as Links from 'apollo-link-core';
+import { ApolloLink, execute } from 'apollo-link-core';
+
+import { createApolloFetch } from 'apollo-fetch';
+
+import { print } from 'graphql';
+import gql from 'graphql-tag';
+import * as fetchMock from 'fetch-mock';
+
+describe('DedupLink', () => {});
```
stubailo: Could use some tests for sure :]

packages/apollo-link-dedup/src/dedupLink.ts
evanshauser: I think it's better to opt-out since the user is including it in their composition, so I switched it up to `forceFetch`. Hoping someone can come up with a better name 🤞

packages/apollo-link-dedup/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "apollo-link-dedup",
+  "version": "0.0.0",
```
jbaxleyiii: lets bump this version so it can be released

packages/apollo-link-dedup/src/dedupLink.ts
jbaxleyiii: @evanshauser I think forceFetch is good especially considering the comment above

PR #34, opened 2017-08-04
apollo-link-batch-http and zen-observable type export

evanshauser:
This adds the batch http link and exports `ZenObservable` typing from `apollo-link-core`


packages/apollo-link-batch-http/src/batching.ts
```diff
@@ -0,0 +1,118 @@
...
+  private apolloFetch: ApolloFetch;
+
+  constructor({
+    batchInterval,
+    batchMax = 0,
+    apolloFetch,
+  }: {
+    batchInterval: number;
+    batchMax?: number;
+    apolloFetch: ApolloFetch;
```
jbaxleyiii: @evanshauser would it be possible to abstract this into something like the `dedupe` link? Where it batches across link chains then plays back the `next` accordingly?

evanshauser: @jbaxleyiii I like that. I think it will be pretty easy to break out the batching into another link. I'll do that with the next commit

PR #63, opened 2017-09-06
return subscriber to it can be unsubscribed

jbaxleyiii:
Dedup link wasn't passing the subscriber so it couldn't be cleaned up 😢 

Fixed it!


stubailo:
noice!

PR #73, opened 2017-09-13
Rename

jbaxleyiii:
This PR removes the bulk `apollo-link` package in favor of renaming `apollo-link-core` to `apollo-link`. Since we are doing an `instanceof` check, we need to ensure only one version of `ApolloLink` is loaded so we moved the dependency of the core lib to a peer dependency.

This is deployed under the `beta` tag for testing across link packages


jbaxleyiii:
cc @evanshauser @peggyrayzis

Issue #76, opened 2017-09-17
Link does not extend ApolloLink and implement request


PCreations:
**Intended outcome:**

I want to create a custom `apollo-link` from a custom build on a custom repo.

**Actual outcome:**

I get the error `Link does not extend ApolloLink and implement request` when trying to use this custom link from outside. The custom link prototype comes from `apollo-link-core` from within its own `node_modules` folder whereas links imported from official repo have their prototype coming from `apollo-link` directly.

**How to reproduce the issue:**

 - Clone this repo : https://github.com/PCreations/apollo-link-logging.git
 - `yarn`
 - `yarn build`
 - `yarn link`

 - Clone this other repo : https://github.com/PCreations/react-apollo-error-template
 - `yarn`
 - `yarn link "apollo-link-logging"`
 - `yarn start`
 - open your browser and open the console
 - the error message should appear

Aetherall:
Same error here

slightlytyler:
@PCreations @Aetherall don't use `apollo-link-core`. instead install `apollo-link@0.6.1-beta.0` directly and import from the there instead. the upgrade docs need to be updated to reflect this

PCreations:
@slightlytyler It doesn't change anything for me. Look at the repos I linked above, I updateed the dependency but still the same problem. Maybe It has something to do with the way I bundle the package with rollup ?

PCreations:
Ok I found the issue I think. Using `instanceof` is not safe, and considered a code smell and harmfull. We shoul favor duck typing instead. `instanceof` can be buggy when dependencies are not correctly linked together and ends up being "loaded" twice (even if it's the same version). This can happen when linking a package when developping locally. That's what I did with a `yarn link` command. Publishing the package to npm solved the issue but we should definitively avoid using `instanceof` I think.

PR #79, opened 2017-09-19
removed instanceof by simple duck typing for testing if provided link…

PCreations:
… is an ApolloLink

It solves the issue #76 by not using `instanceof` anymore and replacing it with simple duck typing test.
This PR is just here to get you the idea and doesn't brake any test but some contributors might want to add a new test case to avoid side effect that I'm not aware of.


jbaxleyiii:
@PCreations thanks so much! I actually changed the validation method totally with https://github.com/apollographql/apollo-link/pull/81

PR #80, opened 2017-09-21
Jest

jbaxleyiii:
This sets up the dev environment around apollo-link to be a 1:1 match with apollo-client including adding danger and filesize tracking


stubailo:
Sounds like a good idea to me!

PR #81, opened 2017-09-21
Improve link API and context

jbaxleyiii:
After writing a number of links on projects, I ran into a few issues that I'm seeking to solve with this PR and a follow up for managing the mutability of the context.

This PR seeks to remove the concept of pure functions for links in favor of allowing the request handler to be the only argument of the constructor for creating a new link. Along the same lines, relying on `abstract` for implementing request doesn't work in non typescript environments so I've opted to throw instead.

After a number of discussions with others on the Apollo team, I realized that forcing the query into an AST node from a string isn't a good idea. Long term, we want to be able to use a query hash id instead of the AST to prevent the need of loading the printer / parser / etc in favor of a system that can identify the hash value and look up the query (ala persisted queries). The current way ApolloLink handled that wouldn't work in this model. Instead query is expected to be a fully formed AST which will become optional in the future if a `queryId` is present. @Poincare is currently working on a persisted query link for example that will use this.

This PR also adjusts how context is used. There are two main issues with context as it stands right now:
1. It is far to easy to clobber the rest of the context and mutate the shared object. In practice I think this could lead to nasty bugs where context is manipulated within an Observable and another Observable may be reading it and getting different results at a given time. 
2. Context isn't accepted by any GraphQL servers so including it as part of the "operation" that gets sent to the server (by thinks like HttpLink, etc) breaks a lot of strongly typed servers (like GraphQLJava). Context can also contain things that can't be serialized so it shouldn't be part of the POST request at all.

With this update, context is on the operation but not directly accessible. Instead there is a `getContext` and `setContext` method to manage the information.

When dealing with stateful links, it is important to isolate the storage of information (like subscriptions) to be per execution. To make this easier, the operation has a `toKey` method which returns a string signature of the operation + variables.


packages/apollo-link/src/link.ts
```diff
@@ -8,179 +10,123 @@ import {
...
-    }
+// XXX add in type support for context manipulation
+// XXX remove mutability of context
+export class ApolloLink {
+  constructor(request?: RequestHandler) {
+    if (request) this.request = request;
   }
 
+  public static empty = empty;
+  // XXX should we deprecate from in favor of fold?
```
stubailo: `fold` makes me feel like I'm doing functional programming - honestly it makes me uncomfortable.

stubailo: Could it be `concat`?

packages/apollo-link/src/types.ts
```diff
@@ -4,16 +4,13 @@ import { ExecutionResult, DocumentNode } from 'graphql';
 export interface GraphQLRequest {
   query?: string | DocumentNode;
   variables?: Record<string, any>;
-  context?: Record<string, any>;
-}
-
-export interface Operation {
-  query: DocumentNode;
```
stubailo: I don't really understand why we don't want the AST here - how could someone write a link that inspects the contents of the query in the new world?

packages/apollo-link/src/link.ts
jbaxleyiii: I'll revert it back to from. I initially thought this would be a problem in case `from` became part of the lexical grammar (with `import Foo from 'bar'`), but for now according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar its not so I think its okay to use.

Agreed on fold

packages/apollo-link/src/types.ts
jbaxleyiii: I think its reasonable to create a version of apollo-client / apollo-link that operations solely on a backend cache to manage updates instead of having the AST on the client.

The client would send an id of the operation + variables and would get back the resulting data and a set of queryIds that need to be invalidated (and possible the new data). The store would be a pretty dumb / low level Map.

Essentially I think that `query` should be a required field, but that it should be `string | DocumentNode`

stubailo: > I think its reasonable to create a version of apollo-client / apollo-link that operations solely on a backend cache to manage updates instead of having the AST on the client.

I feel that this is too far in the future to consider in this version of Link; if we make the implementation too generic that makes it less useful IMO.

> how could someone write a link that inspects the contents of the query in the new world?

Still wondering about this.

evans: I think I agree with James on the move to using a `string` for the `query` field. I was very worried that the move would make splitting on query properties very difficult and a `string` would be too generic.

The positives of having a string is that it makes persistent queries possible without additional effort. We don't have to include the printer/parser for every link stack.

The downside is additional trouble with DocumentNode specific behavior, such as splitting on subscriptions vs mutations/queries. My impression is that this will be a common feature that advanced solutions will need, so adding overhead is okay. The solution is to create a parsing link that adds the DocumentNode and possibly information about the operation type to the context. It also provides people the flexibility to define their own metadata around what a string query does.

Another issue that moving to strings causes is lack of error handling around an invalid query during runtime. This should ideally be done by codegen or some linting software if the queries are static, which leads into the next issue. Using a string means difficulty adding fields to a query, such as `__typename`. Not quite sure how to make this easier yet.

Another option is to make the query an object. This helps with changing fields and introspection. The translation to a network request would be more difficult.

packages/apollo-link-http/src/httpLink.ts
```diff
@@ -26,10 +26,16 @@ export default class HttpLink extends ApolloLink {
 
   public request(operation: Operation): Observable<FetchResult> | null {
     this.headers = (operation.context && operation.context.headers) || {};
+    const { operationName, variables, query, extensions } = operation;
     const request = {
```
evans: This is great!

packages/apollo-link-http/src/httpLink.ts
```diff
@@ -26,10 +26,16 @@ export default class HttpLink extends ApolloLink {
...
-      ...operation,
-      query: print(operation.query),
-    };
+      operationName,
+      variables,
+      query: print(query),
+    } as any;
+
+    if (operation.context.includeExtensions && extensions) {
+      request.extensions = extensions;
```
evans: I was worried about confusing `extensions` and `context`. For now I think it is fine to separate the two into what is sent over the network vs kept in the chain. In the future, if context is sent to the server, the two may become redundant.

docs/source/index.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Apollo Link
+sidebar_title: Introduction
+description: Apollo Link is a standard interface for modifying control flow of GraphQL requests and fetching GraphQL results.
+---
+
+This is the official guide for using Apollo Links in your Apollo app using the [apollo-link](https://github.com/apollographql/apollo-link) package. Apollo Link is a simple yet powerful way to describe how you want to execute a GraphQL operation, and what you want to do with the results. It is a convenient way to customize Apollo Client to the needs of your application while providing a small API to learn.
```
stubailo: Hmm - as a branding thing, should we say "Apollo Links" or "Apollo Link"?

docs/source/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+Follow the repository on GitHub: [Apollo Link](https://github.com/apollographql/apollo-link).
+
+<h2 id="tutorials">Getting started</h2>
+
+If you're familiar with web development, but haven't tried GraphQL or Apollo before, we've got you covered. Here's a set of small tutorials and examples you can look at, and in just a few hours you'll be well on your way to being an expert GraphQL developer! Or, you can jump right in to [installing the library](initialization.html).
```
stubailo: This doesn't really fit, and I think the rest of this page is just the content from the react docs.

docs/source/installation.md
```diff
@@ -0,0 +1,78 @@
+---
+title: Setup and options
+---
+<h2 id="installation">Installation</h2>
+
+To get started with Apollo Link, install the `apollo-link` npm package. This exports everything you need to get started, though there may already be a link with the functionality you wan't on npm.
```
stubailo: > though there may already be a link with the functionality you wan't on npm

I don't think this is super helpful. We should have one canonical way to get started.

docs/source/installation.md
```diff
@@ -0,0 +1,78 @@
...
+Links can be used as a stand-alone client or with most major GraphQL clients.
+
+```js
+import { execute, makePromise } from 'apollo-link';
+import HttpLink from 'apollo-link-http';
+
+const uri = 'http://api.githunt.com/graphql';
+const link = new HttpLink({ uri });
+
+execute(link, operation).subscribe({
```
stubailo: I'd mention that this is the `Observable` interface.

docs/source/installation.md
```diff
@@ -0,0 +1,78 @@
...
+  complete: () => console.log(`complete`),
+})
+
+//For single execution operations, a Promise can be used
+makePromise((execute(link, operation))
+  .then(data => console.log(`received data ${data}`))
+  .catch(error => console.log(`received error ${error}`))
+```
+
+`execute` accepts a standard GraphQL request with an AST (parsed query string using a library like graphql-tag) and returns an [Observable](https://github.com/zenparsing/zen-observable#api) that allows subscribing.
```
stubailo: > accepts a standard GraphQL request with an AST

I'm not sure if I would know what to pass in to `execute` based on this sentence.

docs/source/installation.md
```diff
@@ -0,0 +1,78 @@
...
+  link: new HttpLink({ "/graphql" }),
+  // more options
+});
+```
+
+Under the hood, Apollo Client uses the `execute` function for each operation requested.
+
+<h2 id="graphiql">GraphiQL</h2>
+
+GraphiQL provides a simple way of testing a link.
```
stubailo: I wouldn't say it's just for testing - I think it's a primary use case and a great way to set up GraphiQL!

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Concepts
+---
+<h2 id="concepts">Link Concepts</h2>
+
+Apollo Link is designed to be a powerful way to compose actions around data handling with GraphQL. Each link represents a subset of functionality that can be composed with other links to create complex control flows of data. At a basic level, a link is a function that takes an operation and returns an observable. Described with types, it looks like this:
+
+```js
+type Context = { [key: string]: any };
```
stubailo: I think this is extra for docs, even though it's technically necessary for Flow.

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+<h2 id="concepts">Link Concepts</h2>
+
+Apollo Link is designed to be a powerful way to compose actions around data handling with GraphQL. Each link represents a subset of functionality that can be composed with other links to create complex control flows of data. At a basic level, a link is a function that takes an operation and returns an observable. Described with types, it looks like this:
+
+```js
+type Context = { [key: string]: any };
+
+interface Operation {
+  query: DocumentNode;
+  variables: { [key: string]: any };
```
stubailo: could just be `Object` or `any`, would make the docs more readable IMO

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+  operationName: string;
+  extensions?: { [key: string]: any };
+  getContext(): Context;
+  setContext(newContext: Context | (prevContext: Context) => Context): void;
```
stubailo: I think we could get rid of the function argument API.

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+type RequestHandler = (operation: Operation, forward: NextLink) => Observable<ExecutionResult>
+```
+
+As you can see from these types, the next link is a way to continue the chain of events until data is fetched from some data source (typically a server).
+
+<h3 id="terminating">Terminating Links</h3>
+
+Since link chains have to fetch data at some point, they have the concept of a `terminating` link and `non-terminating` links. Simply enough, the `terminating` link is the one that doesn't use the `forward` argument, but instead turns the operation into the result directly. Typically this is done with a network request, but the possibilities are endless of how this can be done. The terminating link is the last link in the composed chain.
+
+<h3 id="composition">Compostion</h3>
```
stubailo: Composition

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+
+Links are designed to be composed together to form control flow chains to manage a GraphQL operation request. They can be used as middleware to perform side effects, modify the operation, or even just provide developer tools like logging. They can be afterware which process the result of an operation, handle errors, or even save the data to multiple locations. Links can make network requests including HTTP, websockets, and even across the react-native bridge to the native thread for resolution of some kind.
+
+Composition is done using helper functions exported from the `apollo-link` package, or conviently located on the `ApolloLink` class itself. These helpers are explained more [here](./links/composition)
```
stubailo: I think we should pick one to recommend.

docs/source/concepts.md
stubailo: I'd add more detailed docs about the `forward` function here

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+
+<h3 id="context">Context</h3>
+
+Since links are meant to be composed, they need an easy way to send metadata about the request down the chain of links. They also need a way for the operation to send specific information to a link no matter where it was added to the chain. To accomplish this, each `Operation` has a `context` object which can be set from the operation while being written and read by each link. The context is read by using `operation.getContext()` and written using `operation.setContext(newContext)` or `operation.setContext((prevContext) => newContext)`. For example:
```
stubailo: I'd pick just one way to set, and also clarify that `context` is just for links and not sent to the server.

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+
+```js
+const link = new ApolloLink((operation, forward) => {
+  operation.setContext({ start: new Date() });
```
stubailo: In this case, wouldn't a local variable work just as well? I feel that context only make sense for communicating between links.

docs/source/concepts.md
```diff
@@ -0,0 +1,80 @@
...
+  return forward(operation);
+})
+
+const client = new ApolloClient({
+  cache: new InMemoryCache()
+  link,
+});
+
+// send context to the link
+const query = client.query({ query: MY_GRAPHQL_QUERY, context: { saveOffline: true }});
```
stubailo: Is this a thing we support today? This would be a game changer!

docs/source/stateful.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Stateful Links
+---
+
+<h2 id="stateless">Stateless Links</h2>
```
stubailo: Stateful

docs/source/stateful.md
```diff
@@ -0,0 +1,53 @@
...
+
+class OperationTimingLink extends ApolloLink {
+  constructor() {
+    super();
+    this.operations = {};
+  }
+  request(operation, forward) {
+    const key = operation.toKey();
+    if (!this.operations[key]) this.operations[key] = {}
+    this.operations[key].start = new Date();
```
stubailo: Again, this feels like it could be handled with a local variable. I think a dedup link might be more relevant.

packages/apollo-link-batch-http/README.md
```diff
@@ -0,0 +1,59 @@
...
+const client = new ApolloClient({
+  link: new BatchLink({ uri: "/graphql" }),
+  cache: new InMemoryCache()
+});
+
+// a query with apollo-client that will be batched
+client.query({
+  query: MY_QUERY,
+  context: {
+    // example of setting the headers with context per operation
```
stubailo: This seems out of place in the batching docs.

packages/apollo-link-batch-http/README.md
```diff
@@ -0,0 +1,59 @@
+# Batch HTTP Link
```
stubailo: I think we want all of these to be in the docs site - maybe we can symlink so that it's both a README and a docs page?

packages/apollo-link-dedup/README.md
```diff
@@ -0,0 +1,51 @@
+# Dedup Link
+
+## Purpose
+An Apollo Link to deduplicate matching requests before making a request.
+
+*NOTE* This link is included by default when using apollo-client.
```
stubailo: What does this mean? So if I pass in an `HttpLink` then this still gets added to it?

packages/apollo-link-dedup/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Purpose
+An Apollo Link to deduplicate matching requests before making a request.
+
+*NOTE* This link is included by default when using apollo-client.
+
+## Installation
+
+`npm install apollo-link-dedup --save`
+
+To use this package in a web browser or mobile app, you'll need a build system capable of loading NPM packages on the client.
```
stubailo: I don't know if we need this sentence anymore, since more or less everyone is using a build system like this now.

packages/apollo-link-http/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Usage
+```js
+import HttpLink from "apollo-link-http";
+
+const link = new HttpLink({ uri: "/graphql" });
+```
+
+## Options
+HTTP Link takes an object with three options on it to customize the behavoir of the link. If your server supports it, the HTTP link can also send over metadata about the request in the extensions field. To enable this, pass includeExtensions as true
```
stubailo: Any reason we wouldn't want that on by default? GraphQL-Java I guess?

packages/apollo-link-http/README.md
```diff
@@ -0,0 +1,58 @@
...
+|name|value|default|required|
+|---|---|---|---|
+|uri|string|"/graphql"|false|
+|includeExtensions|boolean|false|false|
+|fetch|ApolloFetch|ApolloFetch|false|
+
+By default, this link uses the [Apollo Fetch](https://github.com/apollographql/apollo-fetch) library for the HTTP transport.
+
+## Context
+The HTTP Link uses the `headers` field on the context to allow passing headers to the HTTP request.
```
stubailo: I think we need more options, specifically the `cors` stuff for `fetch`. AFAIK we should have some way to set arbitrary fetch options.

packages/apollo-link-ws/README.md
```diff
@@ -0,0 +1,43 @@
...
+import WSLink from "apollo-link-ws";
+import { SubscriptionClient } from 'subscriptions-transport-ws';
+
+const GRAPHQL_ENDPOINT = 'ws://localhost:3000/graphql';
+
+const client = new SubscriptionClient(GRAPHQL_ENDPOINT, {
+  reconnect: true,
+});
+
+const link = new WSLink(client);
```
stubailo: AFAIK in the new world `client` is already a Link

docs/_config.yml
```diff
@@ -0,0 +1,101 @@
...
+  - '1'
+sidebar_categories:
+  null:
+    - index
+    - installation
+    - concepts
+  Usage:
+    - stateless
+    - stateful
+    - composition
```
stubailo: I think we could really use a `Recipes` section here, with stuff like:

- Setting headers
- Error handling
- Subscriptions

And maybe a `Links` section with all of the individual Link READMEs.

docs/source/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+<h2 id="comparison">Other GraphQL clients</h2>
+
+If you are deciding whether to use `apollo-link` or some other GraphQL client, it's worth considering the [goals](#goals) of the project, and how they compare. Here are some additional points:
+
+ - [Relay](https://facebook.github.io/relay/) is a performance-oriented and highly opinionated GraphQL client built by Facebook for their mobile applications. It focuses on enabling the co-location of queries and components, and is opinionated about the design of your GraphQL schema, especially in the case of pagination. Apollo has an analogous set of features to Relay, but is designed to be a general-purpose tool that can be used with any schema or any frontend architecture. Relay's coupling to a specific architecture enables some benefits but with the loss of some flexibility, which also lets the Apollo community iterate more rapidly and quickly test experimental features.
+ - [Lokka](https://github.com/kadirahq/lokka) is a simple GraphQL Javascript client with a basic query cache. Apollo is more complex, but includes a much more sophisticated cache and set of advanced features around updating and refetching data.
```
evans: How about referencing apollo-fetch instead of Lokka? Both function as thin clients and apollo-fetch is currently maintained

docs/source/installation.md
```diff
@@ -0,0 +1,78 @@
...
+import GraphiQL from 'graphiql';
+
+import { execute } from 'apollo-link';
+import HttpLink from 'apollo-link-http';
+
+const uri = 'http://api.githunt.com/graphql';
+const link = new HttpLink({ uri });
+
+ReactDOM.render(
+  <GraphiQL fetcher={(operation) => execute(link, operation)}/>,
```
evans: Does it make sense to overload execute to have a signature like or just in general make the function curried? 

```js
execute(link: ApolloLink): Operation -> Request

<GraphiQL fetcher={execute(link)}/>,
const result = execute(link)(operation)
```

evans: The `execute(link: ApolloLink): Operation -> GraphQLResult ` constructor can later be extended to include a function that is executed in between links during the `forward` calls.

```js
type Hook = (uplink: ApolloLink, op:Operation, downlink:ApolloLink) -> void

execute(link: ApolloLink, toolingHook:Hook): Operation -> GraphQLResult
```

Certainly wouldn't want this for the first iteration. I could just imagine in the future wanting some way of visualizing how my operation moves through a chain

docs/source/index.md
jbaxleyiii: I think `Apollo Link` since that is also the name of the interface

docs/source/index.md
jbaxleyiii: I thought it would be valuable to copy this if it acts a landing page. I'll remove it for now though!

docs/source/concepts.md
jbaxleyiii: Originally I had but when I was writing the tests it was pretty helpful. It was easier then getting the context then setting it on the next line.

docs/source/concepts.md
jbaxleyiii: Why do you think we should pick just one way? It doesn't seem like much work to maintain both and the function is easier to write when setting context based on previous context while the object argument is easier for simple updates?

docs/source/concepts.md
jbaxleyiii: yeah that's true, I'll adjust the example

docs/source/concepts.md
jbaxleyiii: we do! (not the offline part but the sending context we do!)

packages/apollo-link-batch-http/README.md
jbaxleyiii: even though its in the section about using context to pass headers?

packages/apollo-link-dedup/README.md
jbaxleyiii: yep!

packages/apollo-link-ws/README.md
jbaxleyiii: it has the same request api but isn't actually a link (no split, from, etc). Though since its terminating that shouldn't really matter. I think we should investigate making it a full link though

packages/apollo-link-http/README.md
jbaxleyiii: ah true! I'll take a look. Honestly I think we could / should not use apollo-fetcher here since it duplicates a lot of stuff that could just be handled by a link. All of the `*wares* aren't needed with links. Thoughts?

docs/_config.yml
jbaxleyiii: I agree! I haven't gotten around to doing that yet but we should totally have one! cc @peggyrayzis

docs/source/installation.md
jbaxleyiii: hmm, can this not be done with a link in the chain?

docs/source/concepts.md
stubailo: Dang. It just feels a bit weirdly duplicative.

docs/source/concepts.md
stubailo: I think it makes the docs a bit more complicated - I think maybe it just needs a bit more attention to clarify. I'd maybe present one API as the "normal" one, and the second as an "advanced" option.

docs/source/concepts.md
stubailo: OMG

yessssssssssss

packages/apollo-link-batch-http/README.md
stubailo: Oh wait never mind this is totally legit.

packages/apollo-link-dedup/README.md
stubailo: Do we have any way to override that?

packages/apollo-link-http/README.md
stubailo: I don't have an opinion on this.

docs/source/index.md
stubailo: I think should just remove this section IMO, doesn't make sense to reference Relay here.

packages/apollo-link-http/README.md
evans: The original reason I added apollo-fetch was to create an easy migration path from the current network interface. Not sure that is worthwhile anymore, since we want people entirely on Links.

docs/source/installation.md
evans: Ahh duh you are absolutely right! 

That means we could do `<GraphiQL fetcher={link.request}/>` to reduce boilerplate, if we are willing to suggest the `request` method. Though I remember we talked about avoiding using the `request` method a while ago, so I guess this is bikeshedding now.

So excited to see these changes drop!

packages/apollo-link-http/README.md
jbaxleyiii: @evans totally! I think I'll adjust it and write an upgrade guide!

PR #99, opened 2017-10-02
New http link

jbaxleyiii:
This PR removes the linkage of `apollo-link-http` to `apollo-fetch` in favor of an internal fetch usage. It also has some dev warnings and supports canceling the request if your browser / environment supports it!

cc @evans @peggyrayzis


packages/apollo-link-http/README.md
```diff
@@ -9,28 +9,37 @@ An Apollo Link to allow sending a single http request per operation.
...
 const link = new HttpLink({ uri: "/graphql" });
+
+// or
+import { createHttpLink } from "apollo-link-http";
+
+const link = createHttpLink({ uri: "/graphql" });
 ```
 
+## Global Fetch
+The HTTP Link relies on having `fetch` present in your runtime environement. If you are running on react-native, or modern browsers, this should be no problem. If you are targeting an environement without `fetch` such as older browsers of the server, you will need to pass your own `fetch` to the link through the options. We recommend `unfetch` for older browsers and `node-fetch` for running in node.
```
peggyrayzis: It would be helpful to link to unfetch & node-fetch here

packages/apollo-link-http/README.md
```diff
@@ -9,28 +9,37 @@ An Apollo Link to allow sending a single http request per operation.
...
+
 ## Options
 HTTP Link takes an object with three options on it to customize the behavoir of the link. If your server supports it, the HTTP link can also send over metadata about the request in the extensions field. To enable this, pass includeExtensions as true
```
peggyrayzis: behavior is misspelled here :)

packages/apollo-link-http/README.md
```diff
@@ -53,3 +62,121 @@ client.query({
...
+  operation.setContext({
+    headers: {
+      authorization: localStorage.getItem('token') || null
+    }
+  });
+  return forward(operation)
+})
+
+// use with apollo-client
+const link = middewareLink.concat(httpLink);
```
peggyrayzis: misspelled `middlewareLink`

packages/apollo-link-http/README.md
```diff
@@ -53,3 +62,121 @@ client.query({
...
+```
+
+#### Afterware (error)
```
peggyrayzis: I feel like the error & data manipulation examples are duplicative - maybe move the error example to the error link readme instead? It's a great example, just not sure we need two of them

packages/apollo-link-http/README.md
```diff
@@ -53,3 +62,121 @@ client.query({
...
+
+const httpLink = createHttpLink({ uri: '/graphql' });
+const errorLink = onError(({ networkError }) => {
+  if (networkError.status === 401) {
+    logout();
+  }
+})
+
+// use with apollo-client
+const link = middewareLink.concat(httpLink);
```
peggyrayzis: This should be `errorLink`

packages/apollo-link-http/README.md
peggyrayzis: Also environment is misspelled :)

packages/apollo-link-http/src/__tests__/httpLink.ts
```diff
@@ -215,25 +213,221 @@ describe('HttpLink', () => {
...
+
+    execute(link, { query: sampleQuery, variables }).subscribe(result => {
+      const signal = fetchMock.lastCall()[1].signal;
+      expect(signal).toBe('foo');
+      done();
+    });
+  });
+});
+
+describe('dev warnings', () => {
```
peggyrayzis: WOAH thank you for writing all these tests, next time I will tackle them :) they look great

PR #143, opened 2017-10-12
Fix broken Overview link

thgaskell:
Was linking to `overview.html` instead of `overview.md`.


thgaskell:
Browsing through the rest of the docs, it looks like the markdown files are being compiled for somewhere else? Is there a link to the hosted docs somewhere?

PR #145, opened 2017-10-12
fix(docs): fix typos

ecwyne:


ecwyne:
the request goes forward until it ***GETS*** data

PR #171, opened 2017-10-24
fix: move @types/zen-observable as a dependency instead of a dev one

### LINKED ISSUE #163: @types/zen-observable
I understand why @types/zen-observable is listed as a dev dependency for the apollo-link package, however this may not be ideal as it may cause confusion for some when they cannot build.

As we know, dev dependencies are not transitive, and as such, downstream users utilizing typescript are left having to know to install @types/zen-observable or have their builds fail.

I'll leave this open ended as to a resolution as there are a range of solutions (dependencies, optional dependencies, peer dependencies, better documentation) that may address this better.  Thoughts?

**Intended outcome:**
Expected all necessary files for a typescript build to be installed, so `tsc` would result in a successful build.

**Actual outcome:**
`tsc` results in error TS2688: Cannot find type definition file for 'zen-observable'.

Add the `apollo-link` node module, build using `tsc`

alvis:
Fix #163. All dependent definitions should be tagged as a dependency rather than a dev dependency. All dependent definitions should be tagged as a dependency rather than a dev dependency.

TODO:


jbaxleyiii:
@alvis why should these be dependencies instead of dev?

alvis:
The reason of the need is that if `@types/zen-observable` is declared as a dev dependency, it won't be installed as a dependent when `apollo-link` is used in other project built on top of it. Hence, when we build a higher-up project, typescript will complain for not finding the definitions.
 
See https://github.com/Microsoft/types-publisher/issues/81 for similar issue.

sheeldotme:
@jbaxleyiii see my issue here: https://github.com/apollographql/apollo-link/issues/163#issuecomment-338978391.

schickling:
Is there a new release that already contains this fix?

alvis:
@schickling: not yet. The fix has been merged to the HEAD, and yet a new release is yet to come.

PR #173, opened 2017-10-24
update theme and link readmes to docs

jbaxleyiii:
This should make it way easy to manage the readmes and the docs updates in one place!


jheinnic:
I just noticed the README page for apollo-link-ws change out from underneath me and wanted to pass along an FYI that the Markdown for the page title doesn't seem to be picking up a theme style or something...

It was previously the text "WebSocket Link" in a header-sized font.  

# WebSocket Link

Now it's a to a table with one row containing "WebSocket Link" as the value of a single column entitled "title".  At least that's what appears when browsing the page through GitHub.

title | 
----- |
WebSocket Link |

Curiously, the markdown I supplied above to get this to render as a table isn't what I see in the README.md file itself...  That seems like it should render like follows:

---
title: WebSocket Link
---

PR #180, opened 2017-10-26
Fix createHttpLink typo

Stereobit:
Fixes the function used in the no fetch error message


jbaxleyiii:
Thank you so much! Someone else got to this right before! But thank you!

PR #222, opened 2017-11-08
Fixes GraphiQL Examples

n1ru4l:
Small fix


evans:
It looks like this was solved with #263. Thank you so much for pointing out the issue in the beginning @n1ru4l!

Issue #66, opened 2017-09-07
Mocked Link for Testing


tomitrescak:
This is more a question then issue, but could be part of new features.
I am trying to create a link that would return mocks.

Basically I'm trying to use apollo-test-utils functionality with Apollo Client 2.0. 

would something like this do?

```js
import { ApolloLink, Operation, FetchResult, Observable } from 'apollo-link-core';
import { ApolloFetch, createApolloFetch } from 'apollo-fetch';

import { print } from 'graphql/language/printer';
import { IExecutableSchemaDefinition } from 'graphql-tools/dist/Interfaces';
import { graphql, GraphQLSchema } from 'graphql';

export default class MockLink extends ApolloLink {
  schema: GraphQLSchema;
  rootValue: any;
  context: any;

  constructor(params?: { schema: GraphQLSchema; rootValue?: any; context?: any }) {
    super();
    this.schema = params.schema;
    this.rootValue = params.rootValue;
    this.context = params.context;
  }

  public request(operation: Operation): Observable<FetchResult> | null {
    const request = {
      ...operation,
      query: print(operation.query)
    };

    return new Observable<FetchResult>(observer => {
      graphql(this.schema, request.query, this.rootValue, this.context, request.variables, request.operationName)
        .then(data => {
          if (!observer.closed) {
            observer.next(data);
            observer.complete();
          }
        })
        .catch(error => {
          if (!observer.closed) {
            observer.error(error);
          }
        });
    });
  }
}
```

used like this

```js
const typeDefs = `
Query {
...
}
`;

const mocks = {
    Query: () => ...,
    Mutation: () => ...
  };
  
  const schema = makeExecutableSchema({ typeDefs });
  addMockFunctionsToSchema({
    schema,
    mocks
  });

  const apolloCache = new InMemoryCache(window.__APOLLO_STATE_);

  const graphqlClient = new ApolloClient({
    cache: apolloCache,
    link: new MockLink({ schema }) as any
  });
```

tomitrescak:
Hi, I have tested it and it works rather well. Would you be interested in PR for this? I would call it "LocalLink" instead of "MockLink"

bkniffler:
I think this would be great for server side rendering / querying directly against the schema/database.

tomitrescak:
Yeah, looks like a good use for it. So far no response from devs, I guess they are busy with 2.0 release and once it's done we will get attention. Not in hurry for this.

jbaxleyiii:
[...]

PR #227, opened 2017-11-09
Initial Implementation of LocalLink

tomitrescak:
- [x] Add your name and email to the AUTHORS file (optional)

As discussed in #66.

I could not figure out how to mock graphql import, so one more test is necessary. Docs probably could be improved as well.

### DIFF
--- a/packages/apollo-link-local/README.md
+++ b/packages/apollo-link-local/README.md
@@ -1,0 +1,39 @@
+---
+title: Http Link
+---
+
+## Purpose
+An Apollo Link to allow mocking and server rendering
+
+## Installation
+
+`npm install apollo-link-local --save`
+
+
+## Usage
+```js
+const typeDefs = `
+  Query {
+  ...
+  }
+`;
+
+const mocks = {
+  Query: () => ...,
+  Mutation: () => ...
+};
+
+const schema = makeExecutableSchema({ typeDefs });
+addMockFunctionsToSchema({
+  schema,
+  mocks
+});
+
+const apolloCache = new InMemoryCache(window.__APOLLO_STATE_);
+
+const graphqlClient = new ApolloClient({
+  cache: apolloCache,
+  link: new LocalLink({ schema })
+});
+```
+
--- a/packages/apollo-link-local/package.json
+++ b/packages/apollo-link-local/package.json
@@ -1,0 +1,76 @@
+{
+  "name": "apollo-link-http",
+  "version": "1.1.0",
+  "description": "HTTP transport layer for GraphQL",
+  "author": "Evans Hauser <evanshauser@gmail.com>",
+  "contributors": [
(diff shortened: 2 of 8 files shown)


packages/apollo-link-local/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "apollo-link-http",
```
fbartho: The meta info here probably should be updated!

intellix:
Apologies for my lack of knowledge of how Link, but I'm just wondering why you would need to pass in a schema to return local data? 

I thought the purpose of Link was just for the resolving of the data. In our Angular mocks we were previously doing something like:

```ts
return new ApolloClient({
  networkInterface: {
    query: query => new Promise(resolve => resolve({
      data: {
        hello: 'world',
      },
    })),
  },
});
```

Something like the following seems sufficient for mocking responses to me and was kind of what I was hoping for :)

```ts
import { ApolloLink, Operation, FetchResult, Observable } from 'apollo-link';

export class LocalLink extends ApolloLink {

  constructor(private data: any) {
    super();
  }

    return new Observable<FetchResult>(observer => observer.next(this.data));
  }
}

export default LocalLink;
```

jbaxleyiii:
@tomitrescak this looks great! Could you update the package info and instead of `LocalLink` call it something like `SchemaLink`?

packages/apollo-link-local/README.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Http Link
```
zephraph: Title here should be updated

zephraph:
@jbaxleyiii could you speak a little to what SchemaLink conveys? It seems confusing to me because it seems the other names are how they link or what they do. SchemaLink doesn't quite tell that story, but maybe I'm thinking about it wrong. 

Also, what else needs to be done here? There are a lot of name references that need to be updated (more if the name changes) and it looks like a unit test needs to be finished out. Anything else?

fbartho:
Alternative names?
- ShimLink
- MockLink
- VirtualLink

I think what @jbaxleyiii was getting at with `SchemaLink` was that we're querying, and the Schema itself is generating the response?

zephraph:
I'm fine with the name. You're linking directly to an executable schema, so it makes sense. This can be closed in favor of #284.

zephraph:
@jbaxleyiii we can close this one

imranolas:
Haha, talk about multiple discovery. https://github.com/imranolas/apollo-link-local

stubailo:
Great, closing!

tomitrescak:
Just cam back from a week long conference. Am glad you guys pushed this forward. Great work. @imranolas isn't this now a clone of apollo-link-schema? What is the difference?

imranolas:
@tomitrescak There is no difference hence the deprecation #316

PR #262, opened 2017-11-20
httpLink constructor argument should be optional

bondz:


evans:
@bondz You are absolutely right; the tests definitely check for an empty constructor! Thank you for adding the binding! Merged!

PR #263, opened 2017-11-20
make the GraphiQL example work

staugaard:
Alternative to #222

Just makes the GraphiQL example snippet actually work.


staugaard:
I'm pretty sure the test failure is unrelated

PR #284, opened 2017-11-29
SchemaLink Implementation

zephraph:
- [x] Add your name and email to the AUTHORS file (optional)

As discussed in #66 and started in #227. Credit to @tomitrescak for doing the lion's share of the work here, I just tidied things up a bit.


packages/apollo-link-schema/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "apollo-link-schema",
+  "version": "1.0.0",
+  "description": "Use a GraphQL Schema to request data",
+  "author": "Evans Hauser <evanshauser@gmail.com>",
```
zephraph: Should the author be changed to Tomas since he technically created this one? Just asking, not sure if it matters.

package.json
```diff
@@ -8,8 +8,7 @@
     "test": "lerna run -- test",
     "prelint": "npm run lint-fix",
     "lint": "lerna run -- lint",
-    "lint-fix":
-      "prettier --trailing-comma all --single-quote --write \"packages/*/{src,tests,test,benchmark}/**/*.{j,t}s*\"",
+    "lint-fix": "prettier --trailing-comma all --single-quote --write \"packages/*/{src,tests,test,benchmark}/**/*.{j,t}s*\"",
```
zephraph: I don't know why this was changed exactly. Auto formatting. I can revert it if need be.

packages/apollo-link-schema/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "minify:browser": "uglifyjs -c -m -o ./lib/bundle.min.js -- ./lib/bundle.js",
+    "postbuild": "npm run bundle",
+    "prebuild": "npm run clean",
+    "prepublishOnly": "npm run clean && npm run build",
+    "test": "jest",
+    "watch": "tsc -w -p ."
+  },
+  "dependencies": {
+    "graphql-tools": "^2.7.2",
+    "apollo-link": "^1.0.0"
```
zephraph: I went ahead and moved apollo link to the dependencies like was done in #276

packages/apollo-link-schema/README.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Schema Link
+---
+
+## Purpose
+An Apollo Link to allow mocking and server rendering
```
zephraph: Sound okay? Should this match the description in the package.json?

packages/apollo-link-schema/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Purpose
+An Apollo Link to allow mocking and server rendering
+
+## Installation
+
+`npm install apollo-link-schema --save`
+
+
+## Usage
```
zephraph: We should likely add an SSR example. I'm not 100% sure off hand what that would look like.

zephraph:
@jbaxleyiii It's mostly ready for review. Lots of open questions. I also added a symlinks to the docs to point to this package.

packages/apollo-link-schema/src/__tests__/schemaLink.ts
```diff
@@ -0,0 +1,115 @@
...
+      link,
+    );
+    console.warn = _warn;
+  });
+
+  it('throws if no arguments given', () => {
+    expect(() => new SchemaLink()).toThrow();
+  });
+
+  it('correctly receives the constructor arguments', () => {
```
zephraph: I renamed this test to match its behavior... I don't know how useful it is, but it doesn't hurt so I kept it around.

packages/apollo-link-schema/README.md
zephraph: It solves/codifies this problem.

https://www.apollographql.com/docs/react/recipes/server-side-rendering.html#local-queries

Should we just update those docs to use `SchemaLink` and point it's usage there?

packages/apollo-link-schema/src/schemaLink.ts
```diff
@@ -0,0 +1,57 @@
...
+    context?: any;
+  }) {
+    super();
+
+    this.schema = schema;
+    this.rootValue = rootValue;
+    this.context = context;
+  }
+
+  public request(operation: Operation): Observable<FetchResult> | null {
```
zephraph: `forward` isn't implemented in the request method. My assumption is that not calling forward will break any chaining. Is that correct? 

```js
const consoleLink = new ApolloLink((operation, forward) => {
    // ---- vvvvvvv ---- needs to be preserved?
    return forward(operation).map((data) => { 
      // And this is where the gql call would be made?
      return data;  
    })
})
```

zephraph:
I updated the SSR docs in apollo-client to mention the usage of `SchemaLink` instead of suggesting users write their own. I also updated the local SSR docs...

I think we're good to go. @jbaxleyiii final review when you get the chance.

packages/apollo-link-schema/src/schemaLink.ts
```diff
@@ -0,0 +1,57 @@
...
+    const request = {
+      ...operation,
+      query: print(operation.query)
+    };
+
+    return new Observable<FetchResult>(observer => {
+      graphql(
+        this.schema,
+        request.query,
```
philpl: @jbaxleyiii Hiya, I've written a small addition to the documentation a couple of days ago which coincidentally has been replaced with this.

But I've noticed that you're restringifying the query here instead of using `graphql/execute` and passing through the query AST, which I think can be avoided :smile: (see my docs change) https://github.com/philpl/apollo-client/blame/82a145ef4909a9433ebf3a8c5ef627ceddde79ae/docs/source/recipes/server-side-rendering.md#L233-L236

zephraph: Sorry about stepping on your toes there! Good catch though. I'll put in another PR to correct that.

zephraph: Someone beat me to it. xD

philpl: Yea, no worries :wink: just wanted to let you know. @imranolas is actually my colleague at Formidable who's having a look at Apollo Link and the Apollo Docs too currently 😬

zephraph: Ahh, formidable. <3 you folks.

PR #316, opened 2017-12-06
Deprecate apollo-link-local

imranolas:
...in favor of `apollo-link-schema`. Both packages are practically identical but as `apollo-link-schema` exists under the apollographql org it makes sense to deprecate the former 😢 

TODO:


stubailo:
Oh man! Sorry, I only found out about apollo-link-local yesterday. Is there any difference between the two?

stubailo:
Also I just noticed there aren't docs for the schema link: https://www.apollographql.com/docs/link/

zephraph:
Ahh, sorry @imranolas. Several of us missed the fact your package existed. 

@stubailo, the docs should exist. See https://github.com/apollographql/apollo-link/tree/master/docs/source/links. Schema.md is the link back over to the SchemaLink readme. I'm not sure if I didn't link it correctly or if a special docs deployment needs to happen.

stubailo:
@zephraph I think it needs to be added here: https://github.com/apollographql/apollo-link/blob/master/docs/_config.yml

zephraph:
I see. I'll do a PR for that really quick.

zephraph:
See #319

imranolas:
There's literally no difference. It almost looks like we copy/pasted each others code, haha.

stubailo:
OK, next time we can be more careful to avoid duplicating work! Let's keep writing cool links :]

PR #321, opened 2017-12-07
RetryLink 2.0 - exponential backoff, and split options into delay & retry strategies

### LINKED ISSUE #301: [RetryLink] Simplify by merging delay & interval?
Today, `RetryLink` accepts `delay` and `interval` configuration properties, which seem to have overlapping concerns:

`interval` computes the number of milliseconds until the next retry should begin, and is given the output of `delay`, along with the retry count.  `delay` is only used to determine the input for `interval`, and is given the operation as input.

---

Suggestion, merge them into a single configuration property (`delay` is probably the right name, but maybe something different for backwards compatibility?):

`delay: (operation, count) => ms`

@evans I'm curious what the initial intention was behind the split there?  Am I missing some cases?

nevir:
This introduces a refactored public interface to RetryLink, addressing several pieces of functionality:

* Resolves #301 by allowing `delay` to be configured via a single override

* Introduces exponential backoff by default, including jitter.

* Allows for conditional retry logic (say, for not retrying 4xx class errors)

---

The TL;DR of the new API is:

```ts
new RetryLink({
  delay: {
    initial: 300,
    max: Infinity,
    jitter: true,
  },
  attempts: {
    max: 5,
    retryIf: (_count, _operation, error) => !!error,
  },
});
```

(this is the default configuration - callers should override only the values they care about)

### DIFF
--- a/packages/apollo-link-retry/README.md
+++ b/packages/apollo-link-retry/README.md
@@ -24,17 +24,33 @@ const link = new RetryLink();
 
 ## Options
 
-The standard retry strategy provides exponential backoff with jittering, and takes the following options:
+The standard retry strategy provides exponential backoff with jittering, and takes the following options, grouped into `delay` and `attempt` strategies:
 
-- `initialDelay`: The number of milliseconds to wait before attempting the first retry, with a default of `300`.
+- `delay.initial`: The number of milliseconds to wait before attempting the first retry.
 
-- `maxDelay`: The maximum number of milliseconds that the link should wait for any retry, with a default of `Infinity`.
+- `delay.max`: The maximum number of milliseconds that the link should wait for any retry.
 
-- `maxTries`: The max number of times to try a single operation before giving up, with a default of `5`.
+- `delay.jitter`: Whether delays between attempts should be randomized.
 
-- `jitter`: Whether delays between attempts should be randomized, with a default of `true`.
+- `attempts.max`: The max number of times to try a single operation before giving up.
 
-- `retryIf`: A predicate function that can determine whether a particular response should be retried.  By default, any response with an `error` is retried.
+- `attempts.retryIf`: A predicate function that can determine whether a particular response should be retried.
+
+The default configuration is equivalent to:
+
+```ts
+new RetryLink({
+  delay: {
+    initial: 300,
+    max: Infinity,
+    jitter: true,
+  },
+  attempts: {
+    max: 5,
+    retryIf: (_count, _operation, error) => !!error,
+  },
+});
+```
 
 ### On Exponential Backoff & Jitter
 
@@ -44,18 +60,22 @@ Additionally, with `jitter` enabled, delays are randomized anywhere between 0ms 
 
 These two features combined help alleviate [the thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem), by distributing load during major outages.
 
-### Advanced Mode
+### Custom Strategies
 
-Instead of the options object, you may pass a function that provides full control over whether and how a particular error response should be retried.
+Instead of the options object, you may pass a function for `delay` and/or `attempts`, which implement custom strategies for each.  In both cases the function is given the same arguments (`count`, `operation`, `error`).
+
+The `attempts` function should return a boolean indicating whether the response should be retried.  If yes, the `delay` function is then called, and should return the number of milliseconds to delay by.
 
 ```ts
 import { RetryLink } from "apollo-link-retry";
 
-const link = new RetryLink((count, operation, error) => {
-  // Return false if you want to stop retrying this operation.
-  //
-  // Or, alternatively, return the number of milliseconds that it should be
-  // retried after.
+const link = new RetryLink(
+  attempts: (count, operation, error) => {
+    return !!error && operation.operationName != 'specialCase';
+  },
+  delay: (count, operation, error) => {
+    return count * 1000 * Math.random();
+  },
 });
 ```
 
--- a/packages/apollo-link-retry/README.md
+++ b/packages/apollo-link-retry/README.md
@@ -3,43 +3,59 @@ title: Retry Link
 ---
 
 ## Purpose
-An Apollo Link to allow multiple attempts when an operation has failed. One such use case is to try a request while a network connection is offline and retry until it comes back online. You can configure a RetryLink to vary the number of times it retries and how long it waits between retries through its configuration.
+
+An Apollo Link to allow multiple attempts when an operation has failed, due to network or server errors. `RetryLink` provides exponential backoff, and jitters delays between attempts by default. It does not (currently) support retries for GraphQL errors.
+
+One such use case is to try a request while a network connection is offline and retry until it comes back online.
 
 ## Installation
 
-`npm install apollo-link-retry --save`
+```sh
+npm install apollo-link-retry --save
+```
 
 ## Usage
-```js
+
+```ts
 import { RetryLink } from "apollo-link-retry";
 
 const link = new RetryLink();
 ```
 
 ## Options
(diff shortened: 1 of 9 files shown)


packages/apollo-link-retry/README.md
```diff
@@ -3,43 +3,79 @@ title: Retry Link
 ---
 
 ## Purpose
-An Apollo Link to allow multiple attempts when an operation has failed. One such use case is to try a request while a network connection is offline and retry until it comes back online. You can configure a RetryLink to vary the number of times it retries and how long it waits between retries through its configuration.
+
+An Apollo Link to allow multiple attempts when an operation has failed, due to network or server errors. `RetryLink` provides exponential backoff, and jitters delays between attempts by default. It does not (currently) support retries for GraphQL errors.
+
+One such use case is to try a request while a network connection is offline and retry until it comes back online.
```
helfer: Hey @nevir, I think this is a great PR overall. Many really good improvements! I just thought I'd comment on this particular sentence here because I don't think it's good advice:

I've tried it and it doesn't work well when you're actually offline. It's much better to have an actual offline mode with something like [apollo-link-queue](https://github.com/helfer/apollo-link-queue). It's simpler, more predictable and more reliable.

In case you're curious, here are some of the issues with retrying until a device comes back online:

1. You have to retry many times, causing unnecessary processing every time. The more requests are being retried, the more noise you'll have. It's also not clear how high the max number of retries should be set to make sure the request eventually makes it through. 
2. If you're doing exponential backoff with a high MAX_DELAY to reduce the number of unnecessary requests, you're going to end up with a high delay. That means when you do come back online you might still have to wait up to MAX_DELAY before your requests finally make it to the server. That's bad UX.
3. When your requests finally do make it to the server, they will arrive in a pretty random order. This can also happen when your network isn't unreliable, but it's pretty much guaranteed when you're using a retry link when the device is offline. Although you could use [apollo-link-serialize](https://github.com/helfer/apollo-link-serialize) to mitigate this (which I recommend anyway if your requests depend on a specific ordering).

helfer: Also: a word of caution on retrying requests when the server returned an error (that doesn't explicitly ask you to retry). On several occasions I've witnessed a whole load-balanced system being brought down by a single bug that crashed the server. Here's generally how it went down:

1. Client sends a request that causes the server to hit a bug that crashes the server / blocks the thread / etc.
2. Client eventually gets disconnected from server because the server either crashed or became unresponsive.
3. Client retries failed request and eventually gets rerouted to a new server.
4. Client sends request to new server, crashing that server.
5. Repeat steps 2 - 5 until there are no more servers available. If you're lucky servers will come up quickly enough so that there's always some servers available, but even then you're looking at heavily degraded service.

In general, I would recommend only retrying on network failures (or when the server explicitly asks for a retry), and only automatically retrying a handful of times before requiring user interaction (for example telling the user that the request failed and asking them to press a retry button or simply repeating what they did). People usually don't have the patience to bring down the system because they give up and look for other solutions after a few attempts.

nevir: We've been using this strategy in the wild for quite some time now (monkey patched into AC 0.x), and so far it's given us the best results.  Let me try to address the points from both those comments together:

apollo-link-queue looks pretty neat!  The name is a bit misleading though (it's not actually executing them sequentially once you call `open`, but that's a good thing).  

The problem we've had with approaches like that is that there aren't reliable events for closing/opening the queue.  iOS' reachability framework, Android's equivalent, React Native's, and the web APIs for it are unreliable at best.  We frequently see cases where the device thinks its online but requests consistently fail with L1-5 errors, and vice versa.  

And that's fine, those APIs are more reactive than proactive by design—the intention is more around messaging to a user _why_ requests fail, not to predict the (near) future.

> You have to retry many times, causing unnecessary processing every time. The more requests are being retried, the more noise you'll have. It's also not clear how high the max number of retries should be set to make sure the request eventually makes it through.

If the device is truly offline, this is _very cheap_, and the request will fail immediately (there is no link established to even begin the request).  One caveat in RN: if the request payload is large, sending it over the bridge on each retry to find out _can_ be expensive.

> If you're doing exponential backoff with a high MAX_DELAY to reduce the number of unnecessary requests, you're going to end up with a high delay. That means when you do come back online you might still have to wait up to MAX_DELAY before your requests finally make it to the server. That's bad UX.

Doh, yeah, good catch!  This link could definitely use something similar to the queue link where we can actively ask it to re-evaluate pending requests imperatively 👍 

> When your requests finally do make it to the server, they will arrive in a pretty random order. This can also happen when your network isn't unreliable, but it's pretty much guaranteed when you're using a retry link when the device is offline. Although you could use apollo-link-serialize to mitigate this (which I recommend anyway if your requests depend on a specific ordering).

That's out of scope of a link like this, IMO.  Probably should happen upstream from this link, so that ordering can be truly guaranteed if the caller needs it.

There are also many situations where that behavior is _preferrable_.  Each application has varying levels of (implicit, or explicit) priority for the requests they make.  A messaging application may want messages to not be blocked by status updates, or similar.  (Or, in our case, we want to ensure that status updates for someone's job take precedence over background refreshes and other lower priority queries)

> Also: a word of caution on retrying requests when the server returned an error (that doesn't explicitly ask you to retry). On several occasions I've witnessed a whole load-balanced system being brought down by a single bug that crashed the server.

Yeah, that's one case of the thundering herd problem.  Still, I'd argue (from experience doing large scale systems at Amazon, Google) that it's preferable to retry by default:

* If one bad requests can take down a server process, you're screwed regardless of retries (unless that request is _incredibly_ infrequent).  Even a small number of users will tend to take your systems down for these cases.  Best bet is to fix it at the source.

* Transient errors are _extremely_ frequent by comparison: L1-5 errors, load balancer hiccups, DNS being dumb, routing issues, bad state on servers, etc.  It's very common practice to retry on networking errors and HTTP 5xx class errors as a way to mitigate these cases.  On the flip side, we definitely do not want to retry 4xx class errors for any well behaved system

* However, there is danger retrying mutations by default (which this link implicitly does) - most mutations tend not to be idempotent

Also, the jitter strategy here is employed to smooth out the load over servers and mitigate thundering herd cases where the server might not crash, but instead browns out.  https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ has some background on jitter strategies, and how it affects server-side load (and client request latency)

helfer: These are all good points. As I think I mentioned elsewhere, the right way to do it in my opinion is to have both a retry link and an offline queueing link. The queueing link should be closer to the network than the retry link.

> apollo-link-queue looks pretty neat! The name is a bit misleading though (it's not actually executing them sequentially once you call open, but that's a good thing).

I don't see how the name is misleading. Does queueing imply executing requests serially? To me it only means that requests are put in a queue, and eventually executed. They actually do get passed on to the next link in the same order, so there isn't even any reordering. Maybe apollo-link-buffer would be a better name?

> It's also important, the way AC is architected, for these requests to appear as if they are active until they succeed - particularly for optimistic mutations.

Yep, which is what apollo-link-queue does 🙂 

>[Reodering requests is] out of scope of a link like this, IMO. Probably should happen upstream from this link, so that ordering can be truly guaranteed if the caller needs it.

I agree 100%. I just mentioned it because it's something to keep in mind that people often don't think about and don't run into when developing things locally or under ideal network conditions.

> A messaging application may want messages to not be blocked by status updates, or similar.

apollo-link-serialize has a `serializationKey` option for that, but you probably already know that.

> However, there is danger retrying mutations by default (which this link implicitly does) - most mutations tend not to be idempotent.

I think if you're concerned about that, the only way to really solve it is to include some nonce with every request and try to ensure on the server side that the operation is only executed the first time you see it, because even if you only send it once, it's still possible that your message gets duplicated on the network (depending on the protocol, of course).

PR #340, opened 2017-12-15
Added state docs

peggyrayzis:
Don't merge until Tuesday!

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -22,6 +22,7 @@ sidebar_categories:
     - composition
   Links:
     - links/http
+    - links/state
     - links/error
     - links/context
     - links/retry
@@ -29,7 +30,6 @@ sidebar_categories:
     - links/batch-http
     - links/dedup
     - links/schema
-    - links/state
     - links/community
 github_repo: apollographql/apollo-link
 content_root: docs/source
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -29,6 +29,7 @@ sidebar_categories:
     - links/batch-http
     - links/dedup
     - links/schema
+    - links/state
     - links/community
 github_repo: apollographql/apollo-link
 content_root: docs/source
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -114,3 +114,7 @@ theme: meteor
 ## Docs: http://hexo.io/docs/deployment.html
 deploy:
   type:
+
+marked:
+  gfm: true
+  breaks: false
--- a/docs/source/links/state.md
+++ b/docs/source/links/state.md
@@ -12,8 +12,8 @@ In the past, Apollo users stored their application's local data in a separate
 Redux or MobX store. With `apollo-link-state`, you no longer have to maintain a
 second store for local state. Your Apollo Client cache is your single source of
 truth that holds all of your local data alongside your remote data. To access or
-update your local state, you use GraphQL queries and mutations just like you would
-for data from a server.
+update your local state, you use GraphQL queries and mutations just like you
+would for data from a server.
 
 When you use Apollo Client to manage your local state, you get all of the same
 benefits you know and love like caching and offline persistence without having
@@ -57,7 +57,11 @@ const stateLink = withClientState({
 
 To hook up your state link to Apollo Client, concatenate it to the other links
 in your Apollo Link chain. Your state link should be one of the last links in
-your chain, but before `HttpLink` so local queries and mutations are intercepted before they hit the network. It should also go before [`apollo-link-persisted-queries`](https://github.com/apollographql/apollo-link-persisted-queries) if you are using persisted queries.  Then, pass your link chain to the Apollo Client constructor.
+your chain, but before `HttpLink` so local queries and mutations are intercepted
+before they hit the network. It should also go before
+[`apollo-link-persisted-queries`](https://github.com/apollographql/apollo-link-persisted-queries)
+if you are using persisted queries. Then, pass your link chain to the Apollo
+Client constructor.
 
 ```js
 const client = new ApolloClient({
@@ -98,13 +102,16 @@ const WrappedComponent = graphql(UPDATE_NETWORK_STATUS, {
 What if we want to access our network status data from another component? Since
 we don't know whether our `UPDATE_NETWORK_STATUS` mutation will fire before we
 try to access the data, we should guard against undefined values by providing an
-initial state for our query in our resolver map.
+initial state for our query in our resolver map. Your query resolver will only
+fire on a cache miss, which is why we use it to return an initial state.
 
 ```js
 const stateLink = withClientState({
-  Mutation: { /* same as above */ },
+  Mutation: {
+    /* same as above */
+  },
   Query: {
-    networkStatus: () => false,
+    networkStatus: () => ({ isConnected: false, __typename: 'NetworkStatus' }),
   },
 });
 ```
@@ -170,17 +177,20 @@ functions for each GraphQL object type. You can think of a GraphQL query or
 mutation as a tree of function calls for each field. These function calls
 resolve to data or another function call.
 
-The three most important things to keep in mind about resolvers in
+The four most important things to keep in mind about resolvers in
 `apollo-link-state` are this:
 
-1. The cache is added to the context (the third argument to the resolver) for you so you can write and read data from
-   the cache.
+1. The cache is added to the context (the third argument to the resolver) for
+   you so you can write and read data from the cache.
 2. The resolver should return an object with a `__typename` property unless
    you've overridden the `dataIdFromObject` function to not use `__typename` for
    cache keys. This is necessary for Apollo Client to [normalize the data in the
    cache](/docs/react/basics/caching.html#normalization) properly.
 3. Resolver functions can return a promise if you need to perform asynchronous
    side effects.
+4. Query resolvers are only called on a cache miss. Since the first time you
+   call the query will be a cache miss, you should return any default state from
+   your resolver function.
 
 If any of that sounds confusing, I promise it will be cleared up by the end of
 this section. Keep on reading! 😀
@@ -224,8 +234,8 @@ fieldName: (obj, args, context, info) => result;
    parent field or the `ROOT_QUERY` object in the case of a top-level query or
    mutation. Don't worry about this one too much for `apollo-link-state`.
 2. `args`: An object containing all of the arguments passed into the field. For
-   example, if you called a mutation with `updateNetworkStatus(isConnected: true)`, the
-   `args` object would be `{ isConnected: true }`.
+   example, if you called a mutation with `updateNetworkStatus(isConnected:
+   true)`, the `args` object would be `{ isConnected: true }`.
 3. `context`: The context object, which is shared by all links in the Apollo
    Link chain. The most important thing to note here is that we've added the
    Apollo cache to the context for you, so you can manipulate the cache with
@@ -241,12 +251,13 @@ docs](/docs/graphql-tools/resolvers.html#Resolver-function-signature).
 
 <h3 id="async">Async resolvers</h3>
 
-`apollo-link-state` supports asynchronous resolver functions. These functions can either be `async` functions or ordinary functions that return a Promise. This can be useful
-for performing side effects like accessing a device API. If you would like to
-hit a REST endpoint with your resolver, [we recommend checking out
-`apollo-link-rest`](https://github.com/apollographql/apollo-link-rest) instead,
-which is a more complete solution for using your REST endpoints with Apollo
-Client.
+`apollo-link-state` supports asynchronous resolver functions. These functions
+can either be `async` functions or ordinary functions that return a Promise.
+This can be useful for performing side effects like accessing a device API. If
+you would like to hit a REST endpoint with your resolver, [we recommend checking
+out `apollo-link-rest`](https://github.com/apollographql/apollo-link-rest)
+instead, which is a more complete solution for using your REST endpoints with
+Apollo Client.
 
 For React Native and most browser APIs, you should set up a listener in a
 component lifecycle method and pass in your mutation trigger function as the
@@ -332,9 +343,9 @@ const stateLink = withClientState(
 <h2 id="cache">Updating the cache</h2>
 
 When you manage your local data with Apollo Client, your Apollo cache becomes
-the single source of truth for all your local and remote data. To update and read
-from the cache, you access it via the `context`, which is the third argument
-passed to your resolver function.
+the single source of truth for all your local and remote data. To update and
+read from the cache, you access it via the `context`, which is the third
+argument passed to your resolver function.
 
 <h3 id="write-data">writeData</h3>
 
@@ -378,14 +389,16 @@ const user = {
 
 `cache.writeData` should cover most of your needs; however, there are some cases
 where the data you're writing to the cache depends on the data that's already
-there. In that scenario, you should use [the DataProxy methods](/docs/react/features/caching.html) on the Apollo
-cache, which allow you to pass in a query or a fragment.
+there. In that scenario, you should use [the DataProxy
+methods](/docs/react/features/caching.html) on the Apollo cache, which allow you
+to pass in a query or a fragment.
 
 <h2 id="cache-api">Cache API</h2>
 
 The [Apollo cache API](docs/react/features/caching.html) has several other
-methods to assist you with updating and retrieving data. Here are some common use
-cases where it's best to use the DataProxy methods instead of `cache.writeData`.
+methods to assist you with updating and retrieving data. Here are some common
+use cases where it's best to use the DataProxy methods instead of
+`cache.writeData`.
 
 <h3 id="write-query">writeQuery and readQuery</h3>
 
@@ -473,11 +486,11 @@ const todos = {
 };
 ```
 
-In order to toggle our todo, we need the todo and its status from the cache, which
-is why we call `cache.readFragment` and pass in a fragment to retrieve it. The
-`id` we're passing into `cache.readFragment` refers to its cache key. If you're
-using the `InMemoryCache` and not overriding the `dataObjectFromId` config
-property, your cache key should be `__typename:id`.
+In order to toggle our todo, we need the todo and its status from the cache,
+which is why we call `cache.readFragment` and pass in a fragment to retrieve it.
+The `id` we're passing into `cache.readFragment` refers to its cache key. If
+you're using the `InMemoryCache` and not overriding the `dataObjectFromId`
+config property, your cache key should be `__typename:id`.
 
 To write the data to the cache, you can use either `cache.writeFragment` or
 `cache.writeData`. The only difference between the two is that
@@ -502,8 +515,8 @@ shape of the result; rather, they specify where the data is coming from.
 What's really cool about using a `@client` directive to specify client-side only
 fields is that you can actually combine local and remote data in one query. In
 this example, we're querying our user's name from our GraphQL server and their
-cart from our Apollo cache. Both the local and remote data will be merged together
-in one result.
+cart from our Apollo cache. Both the local and remote data will be merged
+together in one result.
 
 ```js
 const getUser = gql`
@@ -522,9 +535,9 @@ const getUser = gql`
 `;
 ```
 
-Thanks to the power of directives and Apollo Link, you'll soon be able to request
-`@client` data, `@rest` data, and data from your GraphQL server all in one
-query! 🎉
+Thanks to the power of directives and Apollo Link, you'll soon be able to
+request `@client` data, `@rest` data, and data from your GraphQL server all in
+one query! 🎉
 
 <h2 id="examples">Example apps</h2>
 
@@ -553,7 +566,8 @@ love to have you on board as a contributor!
 
 You may have noticed we haven't mentioned a client-side schema yet or any type
 validation. That's because we haven't settled on how to approach this piece of
-the puzzle yet. It is something we would like to tackle soon in order to enable schema introspection and autocomplete with GraphiQL in Apollo DevTools.
+the puzzle yet. It is something we would like to tackle soon in order to enable
+schema introspection and autocomplete with GraphiQL in Apollo DevTools.
 
 Type checking at runtime is problematic because the necessary modules from
 `graphql-js` are very large. Including the modules for defining a schema and
@@ -594,7 +608,9 @@ const WrappedComponent = graphql(
 ));
 ```
 
-What if we could shorten it to something like this, so you don't have to write out the mutation details yourself, but it's still implemented as a mutation under the hood?
+What if we could shorten it to something like this, so you don't have to write
+out the mutation details yourself, but it's still implemented as a mutation
+under the hood?
 
 ```js
 withClientMutations(({ writeField }) => (
@@ -603,7 +619,7 @@ withClientMutations(({ writeField }) => (
 ```
 
 Once we find out how people are using `apollo-link-state`, we can start to write
-helper components for making common mutation and query patterns even easier. These
-components will be separate from React Apollo and will live in another package
-in the `apollo-link-state` repo. If you'd like to help build them, please get in
-touch!
+helper components for making common mutation and query patterns even easier.
+These components will be separate from React Apollo and will live in another
+package in the `apollo-link-state` repo. If you'd like to help build them,
+please get in touch!
--- a/docs/source/links/state.md
+++ b/docs/source/links/state.md
@@ -44,7 +44,7 @@ import { withClientState } from 'apollo-link-state';
 
 const stateLink = withClientState({
   Mutation: {
-    networkStatus: (_, { isConnected }, { cache }) => {
+    updateNetworkStatus: (_, { isConnected }, { cache }) => {
       const data = {
         networkStatus: { isConnected, __typename: 'NetworkStatus' },
       };
@@ -57,7 +57,7 @@ const stateLink = withClientState({
 
 To hook up your state link to Apollo Client, concatenate it to the other links
 in your Apollo Link chain. Your state link should be one of the last links in
-your chain. Then, pass your link chain to the Apollo Client constructor.
+your chain, but before `HttpLink` so local queries and mutations are intercepted before they hit the network. It should also go before [`apollo-link-persisted-queries`](https://github.com/apollographql/apollo-link-persisted-queries) if you are using persisted queries.  Then, pass your link chain to the Apollo Client constructor.
 
 ```js
 const client = new ApolloClient({
@@ -74,8 +74,10 @@ in the cache with our resolver map that we passed into our state link.
 ```js
 const UPDATE_NETWORK_STATUS = gql`
   mutation updateNetworkStatus($isConnected: Boolean) {
-    networkStatus(isConnected: $isConnected) @client {
-      isConnected
+    updateNetworkStatus(isConnected: $isConnected) @client {
+      networkStatus {
+        isConnected
+      }
     }
   }
 `;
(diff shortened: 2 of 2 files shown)


docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+description: Manage your local data with Apollo Client
+---
+
+Managing remote data from an external API is simple with Apollo Client, but
+where do we put all of our data that doesn't fit in that category? Nearly all
+apps need to centralize client-side data from user interactions and device APIs
+somewhere!
+
+In the past, Apollo users stored their application's local data in a separate
+Redux or MobX store. With `apollo-link-state`, you no longer have to maintain a
```
stubailo: or you can use `setState`!

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+second store for local state. Your Apollo Client cache is your single source of
+truth that holds all of your local data alongside your remote data. To access or
+update your local state, you use GraphQL queries & mutations just like you would
+for data from a server.
+
+When you use Apollo Client to manage your local state, you get all of the same
+benefits you know and love like caching, optimistic UI, and persistence without
```
stubailo: I don't think optimistic UI makes sense as a top-level feature here.

Also, I think we should watch out for the word "persistence" being overloaded for persisted queries and storing data persistently... not sure how to solve that one.

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+having to set these features up yourself. 🎉 On top of that, you also benefit
+from the excellent developer experience of Apollo DevTools for painless
```
stubailo: Link to devtools here

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+        networkStatus: { isConnected, __typename: 'NetworkStatus' },
+      };
+      cache.writeData({ data });
+      return data;
+    },
+  },
+});
+```
+
+To hook up your state link to Apollo Client, concatenate it to the other links
```
stubailo: I'd really love to see if we could recommend a single way to combine links, like `from`?

```
import { from } from `apollo-link`;

const link = from([
  stateLink,
  httpLink,
]);
```

Looks a lot clearer to me, but maybe it's just me?

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+`apollo-link-state`, let's dive deeper into the resolver map.
+
+<h2 id="resolver">Resolver map</h2>
+
+Your resolver map is where all the magic happens to retrieve and update your
+local data in the Apollo cache. The resolver map is an object with resolver
+functions for each GraphQL object type. You can think of a GraphQL query or
+mutation as a tree of function calls for each field.
+
+Each field must either resolve to a GraphQL primitive (Int, String, etc) or
```
stubailo: That's not technically correct, you can do:

```
resolvers = {
  Query: {
     obj: () => ({ deep: { object: { here: 'hello' } } })
  }
}
```

Then the query `{ deep { object { here }}}` will work even though it's not a scalar.

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+      name {
+        last
+        first
+      }
+    }
+  }
+`;
+```
+
+For this query, you will need to specify a resolver for `Query.user` in your
```
stubailo: This seems like it has veered into general graphql documentation, especially since I don't see a `@client` directive above.

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+resolver map. If `Query.user` returns an object with a name property that
+corresponds to an object with last and first properties, you do not need to
+specify any additional resolvers. GraphQL takes care of that for you!
+
+The three most important things to keep in mind about resolvers in
+`apollo-link-state` are this:
+
+1. The cache is added to the context for you so you can write & read data from
+   the cache.
+2. The resolver should return an object with a `__typename` property. This is
```
stubailo: It's only necessary if your `dataIdFromObject` function uses `__typename`, so IDK

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+   necessary for Apollo Client to [normalize the data in the
+   cache](/docs/react/basics/caching.html#normalization) properly.
+3. Resolver functions can be asynchronous if you need to perform side effects.
```
stubailo: I'd clarify that async means they return a promise.

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+
+If any of that sounds confusing, I promise it will be cleared up by the end of
+this section. Keep on reading! 😀
+
+<h3 id="async">Resolver signature</h3>
```
stubailo: Probably `async` isn't the right URL hash here

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+
+The signature of a resolver function is the exact same as resolver functions on
+the server built with `graphql-tools`. Let's quickly recap the four parameters
+of a resolver function:
+
+```js
+fieldName(obj, args, context, info) => result
```
stubailo: Maybe add a colon:

```js
fieldName: (obj, args, context, info) => result
```

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+1. `obj`: The object containing the result returned from the resolver on the
+   parent field or the `ROOT_QUERY` object in the case of a top-level query or
+   mutation. Don't worry about this one - you probably won't need to use it for
```
stubailo: You definitely need this argument, you just don't need to worry about the root thing.

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+For most applications, your resolver map will probably be too large to fit in
+one file. To organize your resolver map, we recommend splitting it up by
+feature, similar to the Redux [ducks
+pattern](https://github.com/erikras/ducks-modular-redux). Each feature will have
+its own set of queries, mutations, and fields on its resolver map. Then, you can
+merge all of your separate resolver maps into one object before you pass it to
+`withClientState`.
+
+```js
+import { merge } from 'lodash';
```
stubailo: I'm worried about suggesting people use lodash on the client; is there a simpler merge implementation we can recommend?

docs/source/links/state.md
```diff
@@ -0,0 +1,295 @@
...
+import networkStatus from './resolvers/camera';
+
+const stateLink = withClientState(
+  merge({}, currentUser, cameraRoll, networkStatus),
+);
+```
+
+<h2 id="cache">Updating the cache</h2>
+
+<h3 id="write-data">writeData</h3>
```
stubailo: Let's put these in an `API` section at the bottom maybe?

docs/source/links/state.md
peggyrayzis: Which features would you call out instead?

I'm trying to highlight things that are difficult w/ Redux but easy w/ Apollo - optimistic UI definitely fits that category. Do you think `offline persistence` sounds less ambiguous?

stubailo: > Do you think offline persistence sounds less ambiguous?

Yep!

> I'm trying to highlight things that are difficult w/ Redux but easy w/ Apollo

Yeah for sure, I just don't think optimistic UI for client side data is that interesting. Maybe we can talk about being able to easily attach local state to data you got from the server instead? I think the biggest win is client+server data in one place. OH! also, you can use GraphQL instead of writing janky selectors and it's efficient by default.

docs/source/links/state.md
peggyrayzis: I'm down to switch it to `from`, I agree that we need to be more consistent. Do you think `from` is more intuitive than `ApolloLink.from`? I've always thought importing `concat` & `from` from the `apollo-link` package is weird since concat & from are array methods in JS, but maybe that's just me.

docs/source/links/state.md
peggyrayzis: Would adding a `@client` directive help? Idk the right balance to strike here, I want people who aren't too familiar w/ GraphQL or `graphql-tools` to still be able to feel productive right away.

docs/source/links/state.md
peggyrayzis: Since the default `dataIdFromObject` uses `__typename`, I think this recommendation would apply to most people. I guess we can put a disclaimer like `The resolver should return an object with a __typename property unless your dataIdFromObject function doesn't use __typename for its cache keys.`?

docs/source/links/state.md
peggyrayzis: oops good catch!

docs/source/links/state.md
peggyrayzis: maybe `lodash.merge`?

docs/source/links/state.md
peggyrayzis: Yeah I definitely could have phrased this better.

docs/source/links/state.md
stubailo: I don't care much which one we recommend, but I'd love to recommend exactly one and use it in all examples! I find the static method much more intuitive than `link.concat(link)` but not opinionated about whether it should be attached to `ApolloLink`.

docs/source/links/state.md
stubailo: I think it might be OK if we focus on a target audience of people who already know graphql a bit, but I think this section is fine. Let's add that client directive tho

docs/source/links/state.md
stubailo: oof, that would be rough. Let's keep this for now...

docs/source/links/state.md
stubailo: That is still mega huge. I'm kinda allergic to anything with the word `lodash` in it after fighting it earlier this year. I could write a 10-line replacement for merge and include it in link-state if that helps

docs/_config.yml
```diff
@@ -29,6 +29,7 @@ sidebar_categories:
     - links/batch-http
     - links/dedup
     - links/schema
+    - links/state
```
glasser: In the TOC the name doesn't match the others:

image

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
+---
+title: apollo-link-state
+description: Manage your local data with Apollo Client
+---
+
+Managing remote data from an external API is simple with Apollo Client, but
```
glasser: It looks like the norm in this repo is to have paragraphs that are not line-wrapped in the source, leading to single HTML paragraphs. Line breaks in the source like this appear to lead to `<br>`s and docs that don't flow with window sizes. @stubailo is this correct? maybe should be mentioned in docs-docs

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+where do we put all of our data that doesn't fit in that category? Nearly all
+apps need to centralize client-side data from user interactions and device APIs
+somewhere!
+
+In the past, Apollo users stored their application's local data in a separate
+Redux or MobX store. With `apollo-link-state`, you no longer have to maintain a
+second store for local state. Your Apollo Client cache is your single source of
+truth that holds all of your local data alongside your remote data. To access or
+update your local state, you use GraphQL queries & mutations just like you would
```
glasser: "and"

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+for data from a server.
+
+When you use Apollo Client to manage your local state, you get all of the same
+benefits you know and love like caching and offline persistence without having
+to set these features up yourself. 🎉 On top of that, you also benefit from the
+excellent developer experience of [Apollo
```
glasser: (line break is especially noticeable here)

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+DevTools](https://github.com/apollographql/apollo-client-devtools) for painless
+debugging & full visibility into your store.
```
glasser: "and"

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+      };
+      cache.writeData({ data });
+      return data;
+    },
+  },
+});
+```
+
+To hook up your state link to Apollo Client, concatenate it to the other links
+in your Apollo Link chain. Your state link should be one of the last links in
```
glasser: Hmm, you say "one of the last" and then you put it first. You probably *do* want to put it before HTTP, so maybe you should be more explicit, like "one of the last links in your chain, but it should be before your Http link so that local operations get intercepted before they get to the network" or something?

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+
+What if we want to access our network status data from another component? Since
+we don't know whether our `UPDATE_NETWORK_STATUS` mutation will fire before we
+try to access the data, we should guard against undefined values by providing an
+initial state for our query in our resolver map.
+
+```js
+import { withClientState } from 'apollo-link-state';
+
+const stateLink = withClientState({
```
glasser: I feel like this looks a little too much like wanting to have a second stateLink with different parameters somewhere else, rather than just telling that you need to add a line to the existing one?  Maybe just include something like

```
  Mutation: { /* as above */ }
```

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+  Query: {
+    networkStatus: () => null,
+  },
+});
+```
+
+How do we query the `networkStatus` from our component? Similar to mutations,
+just use a query & the `@client` directive! With Apollo Link, we can combine
```
glasser: "and"

Unless I'm missing something and we consider it valid style to use & for "and" in documentation? Feel free to tell me I'm old fashioned.  But if not, search through the rest of the doc for & too; there are more.

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+
+To retrieve the data in your component, bind your query to your component via
+your favorite Apollo view layer integration just like you normally would. Then,
+React Apollo will attach both your remote & local data to `props.data` while
+tracking both loading & error states. Once the query returns a result, your
+component will update reactively.
+
+```js
+const WrappedComponent = graphql(GET_ARTICLES, {
+  props: ({ data: { networkStatus, articles } }) => {
```
glasser: Is this actually helpful? Do people actually commonly write `props` handlers instead of just taking the default? Maybe it would be more useful to demonstrate this working with no props option but with an ordinary component inside that displays the network status.

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+we don't know whether our `UPDATE_NETWORK_STATUS` mutation will fire before we
+try to access the data, we should guard against undefined values by providing an
+initial state for our query in our resolver map.
+
+```js
+import { withClientState } from 'apollo-link-state';
+
+const stateLink = withClientState({
+  Query: {
+    networkStatus: () => null,
```
glasser: Would `{isConnected: false}` be a better default here? The difference between null and undefined in GraphQL is more subtle to me.

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+Your resolver map is where all the magic happens to retrieve and update your
+local data in the Apollo cache. The resolver map is an object with resolver
+functions for each GraphQL object type. You can think of a GraphQL query or
+mutation as a tree of function calls for each field. These function calls
+resolve to data or another function call.
+
+The three most important things to keep in mind about resolvers in
+`apollo-link-state` are this:
+
+1. The cache is added to the context for you so you can write & read data from
```
glasser: maybe be super explicit here: "the context (ie, the third argument to the resolver)"

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+
+```js
+fieldName: (obj, args, context, info) => result;
+```
+
+1. `obj`: The object containing the result returned from the resolver on the
+   parent field or the `ROOT_QUERY` object in the case of a top-level query or
+   mutation. Don't worry about this one too much for `apollo-link-state`.
+2. `args`: An object containing all of the arguments passed into the field. For
+   example, if you called the field with `networkStatus(isConnected: true)`, the
```
glasser: This matches the example better if you do `updateNetworkStatus` rather than `networkStatus` (I'm not sure what the argument would mean semantically for a query).

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+   by [using `apollo-link-context`](/docs/link/links/context.html).
+4. `info`: Information about the execution state of the query. You will probably
+   never have to use this one.
+
+For further exploration, check out the [`graphql-tools`
+docs](/docs/graphql-tools/resolvers.html#Resolver-function-signature).
+
+<h3 id="async">Async resolvers</h3>
+
+`apollo-link-state` supports asynchronous resolver functions. This can be useful
```
glasser: Somewhere here maybe be explicit that these functions can either be `async` functions or ordinary functions that return a Promise.

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+      cache.writeData({ id: `User:${id}`, data });
+      return data;
+    },
+  },
+};
+```
+
+`cache.writeData` should cover most of your needs; however, there are some cases
+where the data you're writing to the cache depends on the data that's already
+there. In that scenario, you should use the DataProxy methods on the Apollo
```
glasser: link?

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+second store for local state. Your Apollo Client cache is your single source of
+truth that holds all of your local data alongside your remote data. To access or
+update your local state, you use GraphQL queries & mutations just like you would
+for data from a server.
+
+When you use Apollo Client to manage your local state, you get all of the same
+benefits you know and love like caching and offline persistence without having
+to set these features up yourself. 🎉 On top of that, you also benefit from the
+excellent developer experience of [Apollo
+DevTools](https://github.com/apollographql/apollo-client-devtools) for painless
```
glasser: I think it's actually worth having an example somewhere with a screenshot of using `@client` queries from the devtools! It's so cool!  It might be worth mentioning in that section that schema introspection (ie autocomplete) doesn't work on `@client` queries though (w/ link to an issue, if we think that in an ideal world that would work).

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+to provide an initial state. This is why we're returning an empty array in our
+`Query.todos` resolver.
+
+To write the data to the cache, you can use either `cache.writeQuery` or
+`cache.writeData`. The only difference between the two is that
+`cache.writeQuery` requires that you pass in a query to validate that the shape
+of the data you're writing to the cache is the same as the shape of the data
+required by the query. Under the hood, `cache.writeData` automatically
+constructs a query from the `data` object you pass in and calls
+`cache.writeQuery`.
```
glasser: I think it might be nice to say somewhere that the example writes out two different objects: a new object with ID `TodoItem:n` and an updated Todos object.

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+The [Apollo cache API](docs/react/features/caching.html) has several other
+methods to assist you with updating & retrieving data. Here are some common use
+cases where it's best to use the DataProxy methods instead of `cache.writeData`.
+
+<h3 id="write-query">writeQuery & readQuery</h3>
+
+Sometimes, the data you're writing to the cache depends on data that's already
+in the cache; for example, you're adding an item to a list or setting a property
+based on an existing property value. In that case, you should use
+`cache.readQuery` to pass in a query and read a value from the cache before you
```
glasser: Wouldn't it be helpful to have a readData that takes an ID that's analogous to writeData?  Or is that basically what readFragment is (except without you having to specify a query)?

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+the cache, which is why we call `cache.readQuery` to retrieve them.
+`cache.readQuery` will throw an error if the data isn't in the cache, so we need
+to provide an initial state. This is why we're returning an empty array in our
+`Query.todos` resolver.
+
+To write the data to the cache, you can use either `cache.writeQuery` or
+`cache.writeData`. The only difference between the two is that
+`cache.writeQuery` requires that you pass in a query to validate that the shape
+of the data you're writing to the cache is the same as the shape of the data
+required by the query. Under the hood, `cache.writeData` automatically
```
glasser: Are there performance implications here — is automatic construction a trivial or complicated process?

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+          name
+          id
+        }
+      }
+    }
+  }
+`;
+```
+
+Thanks to the power of directives & Apollo Link, you'll soon be able to request
```
glasser: What does "soon" mean here? Does it not actually work yet?

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+`@client` data, `@rest` data, and data from your GraphQL server all in one
+query! 🎉
+
+<h2 id="examples">Example apps</h2>
+
+To get you started, here are some example apps:
+
+* [Todo:](https://github.com/apollographql/apollo-link-state/tree/master/examples/todo)
```
glasser: Don't link the colon (same in next entry)

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+the puzzle yet.
+
+Type checking at runtime is problematic because the necessary modules from
+`graphql-js` are very large. Including the modules for defining a schema &
+validating a request against a schema would significantly increase bundle size,
+so we'd like to avoid this approach. This is why we don't send your server's
+entire schema over to Apollo Client.
+
+Ideally, we'd like to perform type checking at build time to avoid increasing
+bundle size. This is comparable to the rest of the JavaScript ecosystem - for
```
glasser: Use a real em dash here: I think three dashes (---) in Markdown does the trick.

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+While `apollo-link-state` is ready to use in your Apollo application today,
+there are a few enhancements we're looking to implement soon before a v1.0
+release.
+
+We want your experience managing local data in Apollo Client to be as seamless
+as possible, so please get in touch if there's a feature you're looking for
+that's not on this list. Additionally, if any of these topics interest you, we'd
+love to have you on board as a contributor!
+
+<h3 id="type-checking">Type checking</h3>
```
glasser: Might also be a good place to mention schema introspection/devtools

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+      status(text: $text) @client
+    }
+  `,
+)(({ mutate }) => (
+  <button onClick={() => mutate({ variables: { text: 'yo' } })} />
+));
+```
+
+What if we could shorten it to something like this so you didn't have to pass in
+a mutation just to write one field to the cache?
```
glasser: what about:

> What if we could shorten it to something like this, so you don't have to write out the mutation details yourself, but it's still implemented as a mutation under the hood that you can see in the DevTools?

docs/source/links/state.md
stubailo: readFragment is the counterpart that takes an ID, since queries always start at the root query object, which doesn't have an ID. that's maybe not a well known thing though

docs/source/links/state.md
stubailo: I think we don't need to mention that it generates a query, cause that sounds expensive. Actually it's super simple.

glasser: To be honest, that explanation helped it make sense to me.

Although I wonder — maybe instead of making a writeData that's like writeQuery but you don't specify a query, and then having all the examples say "you can do either of these", maybe it just wraps writeQuery with a version of writeQuery where the query argument is optional?  That is probably more confusing than helpful, just spitballing here :)

docs/source/links/state.md
glasser: well that doesn't ring true to me since write has writeData/writeQuery/writeFragment but read only has readQuery/readFragment?

docs/_config.yml
peggyrayzis: Good call, I just merged @stubailo's PR that fixes this issue :)

docs/source/links/state.md
peggyrayzis: this is a result of prettier formatting my markdown, I'm going to figure out how to turn it off

docs/source/links/state.md
peggyrayzis: I love that idea!! I think mentioning the lack of schema introspection would fit well at the bottom section where I talk about typechecking.

docs/source/links/state.md
peggyrayzis: Agreed, that's a much clearer way of phrasing it!

docs/source/links/state.md
peggyrayzis: Hmm, false implies that it's definitely not connected, whereas null implies that we don't know whether it's connected or not. I'm cool with changing it though if we think defining initial state with null is a bad practice.

docs/source/links/state.md
peggyrayzis: just did a find & replace for 22 instances of & 😂 thanks for the catch!

docs/source/links/state.md
peggyrayzis: At MLS, we always wrote custom props handlers, but I'm not sure what other people in the wild do. 🤷‍♀️  The main reason I did the handler was so people could see the loading & error states, but we could probably put that logic inside the component instead.

docs/source/links/state.md
glasser: OK, I guess I don't really get how undefined differs from null in the specific context of GraphQL and React.

docs/source/links/state.md
stubailo: I'm not a fan of having a lot of optional arguments myself, but maybe that would make sense!

docs/source/links/state.md
```diff
@@ -0,0 +1,609 @@
...
+
+Let's look at an example where we're using a GraphQL mutation to update whether
+our network is connected with a boolean flag:
+
+```js
+import { withClientState } from 'apollo-link-state';
+
+const stateLink = withClientState({
+  Mutation: {
+    networkStatus: (_, { isConnected }, { cache }) => {
```
glasser: should have update in its name?

docs/source/links/state.md
peggyrayzis: Good call!

docs/source/links/state.md
peggyrayzis: Honestly I'm not sure haha, it depends on what's going on with `apollo-link-rest`. Server data + client-side data work together in one query though.

docs/source/links/state.md
peggyrayzis: After thinking about this one, this might not be the case for users with different cache implementations. `apollo-cache-hermes` doesn't index every node in the cache, for example.

stubailo:
test
test

hmmm

docs/_config.yml
stubailo: This is fixed in the deploy preview now 👍

docs/source/links/state.md
stubailo: @glasser turns out that when GitHub renders markdown, it actually doesn't respect line breaks. I think we don't need to be super opinionated about how we do it, but we should probably match how GitHub renders stuff.

docs/source/links/state.md
stubailo: Fixed 👍

docs/source/links/state.md
stubailo: Fixed 👍

docs/source/links/state.md
stubailo: I think it's because it's possible to infer what you want when you do `writeData` without a query, but it's not really possible to know what data you want if we had a `readData` without a query.

docs/source/links/state.md
glasser: Yeah I don't care what the source looks like but it's definitely weird as a reader for triple clicking to select lines instead of paragraphs, etc :)

docs/source/links/state.md
glasser: Oh haha, you're saying we can't just call `cache.readData()` and have it magically return the right data :)

PR #361, opened 2017-12-22
Fix apollo-link-state example

mattbretl:
Great work on the new docs for apollo-link-state!  Just a simple correction here: `data` is destructured, so it's not within scope.


stubailo:
Good call! Thanks!

PR #363, opened 2017-12-22
fix(apollo-link-schema): pass context from operation into execute

### LINKED ISSUE #359: apollo-link-schema docs don't explain what it is
I had to read the source of apollo-link-schema to understand what it did. The docs don't explain that it is a terminal link which executes the query against a compiled schema with resolvers.

I also find the name to be confusing. I would expect apollo-link-schema to be something like "provide a copy of the schema language schema so that the client can type-check it before sending requests on the network" or something. The name doesn't at all indicate that it actually executes the query itself. I'd expect this to be called something like apollo-link-direct-execution or apollo-link-local-execution. Or maybe just apollo-link-execute since that's the method that's called!

brabeji:
Setting operation context per SchemaLink instance is insufficient. This PR tries to fix that.

- [X] Make sure all of the significant new logic is covered by tests
- [X] Rebase your changes on master so that they can be merged easily
- [X] Make sure all tests and linter rules pass

Edit by @evans
Fixes #359


packages/apollo-link-schema/src/schemaLink.ts
```diff
@@ -40,7 +40,7 @@ export class SchemaLink extends ApolloLink {
           this.schema,
           operation.query,
           this.rootValue,
-          this.context,
+          this.context || operation.getContext(),
```
brabeji: I'm not entirely sure about this. Would a merge be better?

peggyrayzis: Why do you need to pass in `context` to the SchemaLink constructor? I think using just `operation.getContext()` makes sense here.

brabeji: Yeah, that makes sense and I can remove the constructor param which is the original API. At first, I didn’t want to remove it completely, breaking dependent apps without approval.

packages/apollo-link-schema/src/schemaLink.ts
```diff
@@ -12,25 +12,18 @@ export namespace SchemaLink {
...
-    context?: any;
   }
 }
 
 export class SchemaLink extends ApolloLink {
   public schema: GraphQLSchema;
   public rootValue: any;
-  public context: any;
 
-  constructor({ schema, rootValue, context }: SchemaLink.Options) {
```
brabeji: I realized that this wasn't documented in the first place.

packages/apollo-link-schema/src/schemaLink.ts
```diff
@@ -40,7 +33,7 @@ export class SchemaLink extends ApolloLink {
           this.schema,
           operation.query,
           this.rootValue,
-          this.context,
+          operation.getContext(),
```
brabeji: Sorry to bump this, @peggyrayzis I think this is gtg and It's really holding me back :( Thank you!

zephraph:
@jbaxleyiii Would you have a moment to take a look at this? A user pinged me on twitter that was having an issue that I believe this would help resolve.

lzl:
@zephraph Thank you for mention this PR on Twitter.

My original tweet is [here](https://twitter.com/lzl/status/964543936868970497).

> Is there a way to perform SSR with login token & avoid making network calls? I tried to do with authLink and schemaLink, but it seems not work. Need help. 💡 @stubailo @jbaxleyiii @peggyrayzis @zephraph
>
> 

After read the README.md and this PR, I understood what the `context` means in the option of `SchemaLink`.

So I resolved my problem with code:

```js
// copy getUserForContext from https://github.com/apollographql/meteor-integration/blob/2881dc449a1d4e96b814c35ab50f22ba92cdf563/src/main-server.js#L127
const userContext = await getUserForContext(context.loginToken)

const schemaLink = new SchemaLink({
  schema,
  context: userContext,
})

  ssrMode: true,
  link: schemaLink,
  cache: new InMemoryCache(),
})
```

When this PR got merged, code will be:

```js
const userContext = await getUserForContext(context.loginToken)

const authLink = new ApolloLink((operation, forward) => {
  operation.setContext(() => userContext)
  return forward(operation)
})

const schemaLink = new SchemaLink({ schema })

  ssrMode: true,
  link: from([authLink, schemaLink]),
  cache: new InMemoryCache(),
})
```

I found out `meteor/apollo` package did not work with `apollo-link-schema` out of box. Because of `getUserForContext` is part of middleware of `WebApp`, and `apollo-link-schema` does not make network calls. Maybe there is room for improvement at `meteor/apollo` part.

packages/apollo-link-schema/src/schemaLink.ts
zephraph: It wasn't documented but it was provided. This is technically a breaking change. Could it support both cases initially? If not, I think it should be a major bump.

zephraph: On second thought, I think supporting both cases would be more confusing than it's worth. I'm 👍 for this change, but I still think it should be a major bump.

zephraph:
@brabeji when you get a chance could you merge master into this branch? I pinged the maintainers chat so hopefully we can get a conversation going on Monday.

evans:
@brabeji Thank you for the changes! I'm taking a look at it. Since it's a breaking change and `apollo-link-schema` is new to me, it will take me a bit to get spun up on the project

brabeji:
Thank you guys! I appreciate you being careful about breaking changes. I've considered support for constructor param with some sort of merge/preference with/over operation context (yes, is't unnecessarily complex), but then came to the same conclusion as @zephraph. Constant context can easily be achieved with link composition.

Congrats on the ownership @evans!

evans:
I really appreciate you @brabeji for bringing up the original issue and the PR! After thinking about it, I believe the right thing to do is change the constructor signature to allow context to be a function that accepts the link context and returns the resolver context in addition to allowing a static resolver context(following a similar pattern to setting the uri in [link-http](https://github.com/apollographql/apollo-link/blob/0928d36a63a744c8555c76742bdfb1d455889781/packages/apollo-link-http/src/httpLink.ts#L21)).

The unfortunate side effect of `link-schema` using context as a named field in the constructor argument is that it is confused with the link context. I think we should be explicit about the difference, which appears to be a docs issue currently(ideally `context` would be named `resolverContext`). I'm not sure we need to make this a breaking change, unless it makes sense never to have a static resolver context/we want it to be highly discouraged. I'm curious about your thoughts and absolutely open to revising my opinion.

> [Resolver context](http://graphql.org/learn/execution/#root-fields-resolvers) isn't well documented in the [schema link docs](https://www.apollographql.com/docs/link/links/schema.html), so we should document it in this change.

evans:
@brabeji Thank you for bring up the original feature and starting the PR!

@zephraph Thanks for looking over it initially!

evans:
@brabeji we published in apollo-link-schem@1.1.0 🎉 Thanks for the help!

Issue #298, opened 2017-12-02
Subscribing to an observable causes duplicate HTTP requests


joshjg:
**Intended outcome:**
I tried following the pattern of the third example here, and composing with an HttpLink:
https://www.apollographql.com/docs/link/stateless.html

**Actual outcome:**
All queries result in duplicate http requests being sent to the server.

**How to reproduce the issue:**
```js
const httpLink = new HttpLink({ uri: foo });
const errorLink = new ApolloLink((operation, forward) => {
    const observer = forward(operation);
    observer.subscribe({ error: console.error });
    return observer;
});

const link = ApolloLink.from([
    errorLink,
    httpLink
]);
```

jbaxleyiii:
@joshjg hmm yes I think those are written incorrectly in the docs. It should be within an Observable and bound to notify it. I can update those docs, or if you feel up to it, the `apollo-link-error` shows the correct way to write something like that. Would you be able to update the docs matching?

nicocrm:
Picking up on this older issue.  It has been classified as a documentation problem, but it makes the observable that is returned from the HTTP link error-prone to use in custom links.  It would be great if it could be enhanced so that the pattern that is described in the documentation is possible.  Ideally it would take care to only send the request once, even if subscribed to more than once.

Is there any plan to do that, or might you accept a pull request that would add that feature to the HTTP link?  Or am I missing something that in fact does require the link to send the request every time the observable is subscribed to?

evans:
@nicocrm Take a look at #364. At least for batching, we will support sending the request only once per batch regardless of the number of subscribers(see [here](https://github.com/apollographql/apollo-link/blob/http-batching/packages/apollo-link-batch/src/batching.ts#L74-L83)).

I totally agree that we should only send the request once for subscriptions that arrive in time(before and during the fetch call). We'll probably have to make some changes to how we deal with the result of the [fetch call](https://github.com/apollographql/apollo-link/blob/http-batching/packages/apollo-link-http/src/httpLink.ts#L71).

[...]

PR #364, opened 2017-12-23
Share common http functionality between links and update apollo-link-batch-http

### LINKED ISSUE #231: BatchHttpLink does not populate response headers
**Intended outcome:**
I am attempting to add an afterware to a BatchHttpLink in order to see if a header was sent in the response. If I use just an HttpLink I am able to access the expected response header. All the code stays the same, just changing between BatchHttpLink and HttpLink. Here is the code I am using (the batch portions commented out right below the regular http portions). This is on Nuxt also in case it matters for debug:

```
import { ApolloLink } from 'apollo-link'
import { HttpLink } from 'apollo-link-http'
// import { BatchHttpLink } from 'apollo-link-batch-http'
import { InMemoryCache } from 'apollo-cache-inmemory'

export default ctx => {
  const httpLink = new HttpLink({ uri: process.env.GRAPHQL_URL })
  // const httpLink = new BatchHttpLink({ uri: process.env.GRAPHQL_URL })

  // auth token
  let token = 'asdf'

  // middleware
  const middlewareLink = new ApolloLink((operation, forward) => {
    if (token) {
      operation.setContext({
        headers: { authorization: `Bearer ${token}` },
      })
    }
    return forward(operation)
  })

  // refresh token afterware
  const afterwareLink = new ApolloLink((operation, forward) => {
    return forward(operation).map(response => {
      const context = operation.getContext()
      const { response: { headers } } = context
      if (headers) {
        const refreshToken = headers.get('Refresh-Token')
        if (refreshToken) {
          console.log('refresh token found: ', refreshToken)
        }
      }

      return response
    })
  })

  const link = ApolloLink.from([middlewareLink, afterwareLink, httpLink])
  return {
    link,
    cache: new InMemoryCache(),
  }
}
```

**Actual outcome:**
response object is not present at all, therefore unable to read headers property of undefined:
```
Error: Network error: Cannot read property 'headers' of undefined
    at new ApolloError (ApolloError.js:34)
```

### LINKED ISSUE #343: Batching and credentials with apollo-link-batch-http
What about intuitive API for BatchHttpLink like the one in [example](https://www.apollographql.com/docs/react/basics/network-layer.html#) of HttpLink?

Seems to me it should look similar with BatchHttpLink to prevent dumb errors:
```js
import ApolloClient from 'apollo-client';
import { BatchHttpLink } from 'apollo-link-batch-http';

const batchLink = new BatchHttpLink({
  uri: 'https://example.com/graphql',
  // Additional fetch options like `credentials` or `headers`
  credentials: 'same-origin',
});

  batchLink,
  // other options like cache
});
```

evans:
This PR contains code to remove the logic shared across all http links from apollo-link-http and put it into a publicly accessible `apollo-link-http-common` module. This module is used by apollo-link-batch-http to have consistent behavior around errors and serialization.

This PR also add the ability to change a query's batch with the `batchKey` constructor option. `batchKey` is a function that accepts an operation and returns a string key that uniquely identifies the batch it belongs in.

These changes also modify the way that the fetch option and request body are created, found in the `createOptionsAndBody` method (CC @jaydenseric). The config precedence remains the same(least to greatest): default, link constructor, context.

This PR found and fixes a bug in apollo-link-batch for multiple subscribers on a single request. Issues of multiple subscription calls(#298) are still present in apollo-link-http.

TODO:

Fixes #231 Fixes #343


stubailo:
> If this is not the case and you need batching with multiple uri's, please mention it!

I think using the `split` link will handle that just fine, right?

Thanks for getting this started!

AdamYee:
> If this is not the case and you need batching with multiple uri's, please mention it!

This might be a naive question @evans @stubailo, but wouldn't schema stitching be preferred over using a `split` in this case?

packages/apollo-link-batch/src/batchLink.ts
```diff
@@ -58,6 +58,10 @@ export class BatchLink extends ApolloLink {
       batchMax: this.batchMax,
       batchHandler: fetchParams.batchHandler,
     });
+
+    if (fetchParams.batchHandler.length <= 1) {
+      this.request = operation => this.batcher.enqueueRequest({ operation });
```
AdamYee: Is overriding `request` in the constructor a common thing? Instead of replacing `request` here, would it be better to check `fetchParams.batchHandler.length <= 1` inside the `public request` method?

packages/apollo-link-batch/src/__tests__/batchLink.ts
```diff
@@ -268,21 +281,34 @@ describe('OperationBatcher', () => {
...
       expect(batcher.queuedRequests.length).toBe(0);
+      expect(operation.getContext()).toEqual({ response: { data } });
       done();
     }, 20);
   });
 
   it('should correctly batch multiple queries', done => {
+    const data = {
+      lastName: 'Ever',
+      firstName: 'Greatest',
```
stubailo: hotline bling

packages/apollo-link-batch/src/batchLink.ts
stubailo: I'm a bit confused as to why we need to replace the `request` function here at all - what is the `batchHandler.length` measuring? Like the number of arguments to `batchHandler`?

packages/apollo-link-batch/src/batching.ts
```diff
@@ -102,23 +115,40 @@ export class OperationBatcher {
...
+        if (!Array.isArray(results)) {
+          results = [results];
+          if (nexts.length != 1)
+            console.warn(
+              `server returned single result, expected array of length ${
+                nexts.length
+              }`,
+            );
+        }
+        // console.log(results);
```
stubailo: We probably don't want these in the PR

packages/apollo-link-batch/src/batching.ts
```diff
@@ -102,23 +115,40 @@ export class OperationBatcher {
 
     batchedObservable.subscribe({
       next: results => {
+        if (!Array.isArray(results)) {
+          results = [results];
+          if (nexts.length != 1)
+            console.warn(
+              `server returned single result, expected array of length ${
```
stubailo: Hmm should we have a general thing that makes sure the number of nexts equals the number of results? Looking below, it looks like we don't do anything useful in that case right now.

packages/apollo-link-http/src/httpLink.ts
```diff
@@ -179,73 +302,66 @@ export const createHttpLink = (linkOptions: HttpLink.Options = {}) => {
...
-        } catch (e) {
-          const parseError = new Error(
-            `Network request failed. Payload is not serializable: ${e.message}`,
-          ) as ClientParseError;
-          parseError.parseError = e;
-          throw parseError;
+  //currently all operations in a single batch use the same uri
+  let choosenURI: string;
+
+  return new BatchLink({
```
stubailo: Hmmm, I think I would have been expecting something that doesn't go through the batching code path at all unless batching is being used. This looks like it depends on what is passed into the BatchLink constructor, but maybe we could just avoid calling that?

I could imagine a situation where you do something like

```
ApolloLink.from([
  batchLink,
  httpLink
]);
```

or maybe that's wishful thinking

packages/apollo-link-batch/src/batchLink.ts
evans: Yep it's a count of how many arguments the function has. It's how we determine if a link is terminating or non-terminating(based on if request accepts a forward or not).

Admittedly it is a pretty poor way of doing termination/non-termination in this case. If there are better ways to determine if a link is terminating or modifying the public methods signature other than overriding, please share!

packages/apollo-link-http/src/httpLink.ts
evans: Yeah @jaydenseric had similar sentiments! I think that `batchLink.concat(httpLink)` would complicate the request interface and make visual debugging(dev tools tracing a request through the chain) more difficult.

Whether we add the batching behavior to link-http or keep it in link-batch-http is a philosophical question to answer. I'm fine with having the terminating(transport) links owning batching, since I hope that batching becomes a standard. Then a developer will only have to pick their protocol: ws, ddp, http, etc. Adding and maintaining a separate batch version seems like a pain.

packages/apollo-link-batch/src/batching.ts
evans: Ah duh, thank you!

I think the reasonable thing to do is propagate an error with the `response` field of the Error set to `results`.

evans:
I think we can go in a better direction here. @jbaxleyiii pointed out [this article](https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html) in react's blog. The important part is that we can create a shared unpublished package that includes the functionality shared across `apollo-link-http`, `apollo-link-batch-http`, and `apollo-link-upload`(+`apollo-link-batch-upload`?). This private package will be bundled into the source for each package individually, so each of the packages stay in sync. This is a much better solution for the link ecosystem, especially since the [currently proposed code](https://github.com/apollographql/apollo-link/pull/364/files/7d62c6e27ab14435b516a4b3f4e2c424cd9a0c1a) doesn't really do the right thing for batching(by uri with a `Map<String, Batch>`)

CC @jaydenseric

jaydenseric:
For DX it would be best if there was one terminating link that does fetching. No need to consumers to change their dependencies just to turn on batching or begin uploading files. Less packages to maintain and version, and less documentation to maintain and learn. For this approach there is no needs for seperate helpers.

I'm not following why the one package idea is no longer feasible?

If there has to be seperate packages, it would be better to publish a helpers package so it can benefit packages outside of the repo.

evans:
For further context on ^, the current situation is that uri can be generated from the each operation’s `context.uri` or the constructor’s `uri: Operation -> String`. The correct action is to respect these per operation uris and create a batch for each of them that are all fired off on the interval or send an individual batch when it hits `batchMax`. This would add a lot of complexity to the http link, so it is better as a batch http link.

darren128:
What's the status on this?

evans:
@darren128 Thank you for the checkin! I've finished up the code changes. Now we need to add some tests for apollo-link-utilities and the new [`batchKey`] function for `apollo-link-batch` and `apollo-link-batch-http`. Could you add some?

We'll also need to update the documentation.

packages/apollo-link-utilities/README.md
```diff
@@ -0,0 +1,6 @@
+---
+title: apollo-link-utilites
+description: Http utilities shared across Apollo Links
+---
+
+This repository is used in apollo-link-http and apollo-link-http-batch. The package is public to allow easier development of links that will be a part of the main repository. Developers using this package should know that the api will change and versions may not follow SemVer.
```
jaydenseric: Why won't `apollo-link-utilites` versions follow semver?

packages/apollo-link-utilities/src/index.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  const defaultHttpOptions: HttpOptions = {
+    includeQuery: true,
+    includeExtensions: false,
+  };
+
+  const defaultHeaders = {
+    // headers are case insensitive (https://stackoverflow.com/a/5259004)
+    accept: '*/*',
```
jaydenseric: I know it has always been `*/*`, but shouldn't it really be `application/json`, since that is what is required to work?

packages/apollo-link-utilities/src/index.ts
```diff
@@ -0,0 +1,269 @@
...
+        return result;
+      })
+  );
+};
+
+export const checkFetcher = (fetcher: GlobalFetch['fetch']) => {
+  warnIfNoFetch(fetcher);
+};
+
+const warnIfNoFetch = fetcher => {
```
jaydenseric: Could this be merged into `checkFetcher` above, as it's not used anywhere else?

packages/apollo-link-utilities/src/index.ts
```diff
@@ -0,0 +1,269 @@
...
+  if (!fetcher && typeof fetch === 'undefined') {
+    let library: string = 'unfetch';
+    if (typeof window === 'undefined') library = 'node-fetch';
+    throw new Error(
+      `fetch is not found globally and no fetcher passed, to fix pass a fetch for
```
jaydenseric: I think we should get rid of checking fetch and instead document that it needs to be polyfilled in readme setup instructions. This error message can't be minified and adds to the bundle size. The message recommending `node-fetch` might not make sense in a universal codebase: `isomorphic-unfetch` would be better.

jaydenseric: https://www.npmjs.com/package/invariant might be an option, if we retain dev errors.

packages/apollo-link-utilities/README.md
```diff
@@ -0,0 +1,6 @@
+---
+title: apollo-link-utilites
```
jaydenseric: Typo: `utilities`.

packages/apollo-link-http/package.json
```diff
@@ -42,7 +42,8 @@
     "watch": "tsc -w -p . & rollup -c -w"
   },
   "dependencies": {
-    "apollo-link": "^1.1.0"
+    "apollo-link": "^1.1.0",
+    "apollo-link-utilities": "0.0.0"
```
jaydenseric: The `apollo-link-utilities` dependency has been added, but it does not seem to be used yet in `apollo-link-http`?

jaydenseric: My mistake! The diff was hidden.

Skaronator:
@jaydenseric Please consider next time to make a single review instead of adding 8 single comments.

image

Otherwise I'll get a email for every single comment:
image

Thanks

packages/apollo-link-utilities/src/index.ts
```diff
@@ -0,0 +1,247 @@
...
+}
+
+export interface ConfigOptions {
+  http?: HttpOptions;
+  options?: any;
+  headers?: any; //overrides headers in options
+  credentials?: any;
+}
+
+export namespace LinkUtils {
```
jaydenseric: Would it not be better to export everything in this `LinkUtils` namespace separately? At the moment we can't use a named import to just grab what we want (`fallbackConfig` is all I need for `apollo-upload-client`), defeating tree-shaking. `LinkUtils` is also not a very intuitive named import, looking at it you might think it is some sort of class that can be constructed, or that it contains everything in `apollo-link-utilities`. `Utils` sounds like helper functions, but the namespace contains a lot of default options objects.

evans: That makes sense. I wasn't aware of the lack of tree-shaking. Thanks!

You also bring up a good point about the naming. I think there is a general issue in the utilities, so I am going to make the utility functions a bit more descriptive.

packages/apollo-link-http/README.md
```diff
@@ -11,23 +25,21 @@ import { createHttpLink } from "apollo-link-http";
...
+* `uri`: the URI key is a string endpoint or function resolving to an endpoint -- will default to "/graphql" if not specified
 * `includeExtensions`: allow passing the extensions field to your graphql server, defaults to false
 * `fetch`: a `fetch` compatiable API for making a request
 * `headers`: an object representing values to be sent as headers on the request
 * `credentials`: a string representing the credentials policy you want for the fetch call
 * `fetchOptions`: any overrides of the fetch options argument to pass to the fetch call
 
+<h2 id="fetch">Fetch polyfill</h2>
+
+The HTTP Link relies on having `fetch` present in your runtime environment. If you are running on react-native, or modern browsers, this should be no problem. If you are targeting an environment without `fetch` such as older browsers of the server, you will need to pass your own `fetch` to the link through the options. We recommend [`unfetch`](https://github.com/developit/unfetch) for older browsers and [`node-fetch`](https://github.com/bitinn/node-fetch) for running in Node.
```
glasser: "or" the server

jaydenseric:
Just published [`apollo-upload-client@7.1.0-alpha.1`](https://github.com/jaydenseric/apollo-upload-client/releases/tag/v7.1.0-alpha.1), making use of `apollo-link-http-common@0.0.0`.

jbaxleyiii:
@evans I think this is good to go *except* the removal of the runtime warnings like the one for missing fetch. Why were those removed?

evans:
@jbaxleyiii The original reason was to reduce the bundle size. Also since apollo-boost doesn't accept a fetch argument, I thought it was unnecessary. After running into it internally, I agree with you that dev warnings still should be a part of the ecosystem.

evans:
We'll be publishing after #489!

@jaydenseric @AdamYee Thank you for the code review!
@glennreyes Thank you for the tests!

albinekb:
@evans can't you release this without #489 ? Would be nice to start testing it now. Or is there any way to install packages directly from github? I can't seem to figure out how to run code directly from a commit with this monorepo structure. Pointers?

Issue #365, opened 2017-12-23
Add rollup watch to package.json's watch


evans:
Add [rollup-watch](https://github.com/rollup/rollup-watch) as a dev dependency and change the `npm run watch` commands to include `rollup -c -w`. Each package depends on the rollup output, so both tsc and rollup are needed. 

Example can be seen [here](https://github.com/apollographql/apollo-link/pull/364/files#diff-eb600a1819c46a5426b3c9478ba37e67R42).

PR #367, opened 2017-12-23
Add rollup to watch scripts

evans:
This is a build tool addition to add rollup to the incremental compilation, since the links depend on the rollup output, not just the tsc result.

Run `npm run watch` in the root of the repository after an `npm run build`. The `npm run build` is necessary to ensure the earlier link dependencies exist, such as `apollo-link-core`.

FYI @jbaxleyiii @peggyrayzis


evans:
Fixes #365

PR #372, opened 2017-12-27
fix(link-schema): return correct promise [#371]

### LINKED ISSUE #371: [Apollo-link-schema] Network error: graphql.execute(...).then is not a function
**Intended outcome:**

return a correct Promise

**Actual outcome:**

```cmd
networkError: TypeError: graphql.execute(...).then is not a function
          at /code/.../node_modules/apollo-link-schema/lib/bundle.umd.js:31:18
```

evenchange4:
fixed #371 

TODO:

### DIFF
--- a/packages/apollo-link-schema/src/schemaLink.ts
+++ b/packages/apollo-link-schema/src/schemaLink.ts
@@ -35,16 +35,17 @@ export class SchemaLink extends ApolloLink {
 
   public request(operation: Operation): Observable<FetchResult> | null {
     return new Observable<FetchResult>(observer => {
-      Promise.resolve(
-        execute(
+      new Promise(resolve => {
+        const data = execute(
           this.schema,
           operation.query,
           this.rootValue,
           this.context,
           operation.variables,
           operation.operationName,
-        ),
-      )
+        );
+        resolve(data);
+      })
         .then(data => {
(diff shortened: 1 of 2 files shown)


packages/apollo-link-schema/src/schemaLink.ts
```diff
@@ -35,16 +35,17 @@ export class SchemaLink extends ApolloLink {
 
   public request(operation: Operation): Observable<FetchResult> | null {
     return new Observable<FetchResult>(observer => {
-      Promise.resolve(
-        execute(
+      new Promise(resolve => {
```
jedwards1211: @evenchange4 are you sure something is not wrong with your JavaScript environment, like are you using a weird polyfill for `Promise` or something?  [`Promise.resolve`[(http://devdocs.io/javascript/global_objects/promise/resolve) is now standard so I don't see why this change would be necessary, I think either your build or runtime environment isn't configured with a proper `Promise.resolve` implementation.

evenchange4: You are right. There is some problem with my jest environment.

jedwards1211: Ah don't worry, I'm not a maintainer, I just happened to notice this.

PR #380, opened 2018-01-02
Use rxjs instead of zen-observable

evans:
We should have a discussion if rxjs is a good idea, since there are some considerations:

First, rxjs would need to be a peerDependency, which means that we'd have to do similar things to `graphql`, namely add it to our npm install list. The other route would be a dependency for every link that uses any of rxjs functionality outside the base `Observable`. From what I understand this would mean duplicate instances of rxjs code in a bundle. I think it is reasonable to assume that most developers are using some sort of tree shaking already, since the `graphql` dependency almost requires it([background](https://github.com/graphql/graphiql/issues/622)), so I believe that adding rxjs wouldn't require a change to builds. Also note, the bundle size of `apollo-link-batch` is a good example of what adding an operator might look like and the effect on bundle size, though there are [better ways](https://github.com/apollographql/apollo-link/commit/96398279129cbaa98852594e99c8e2c6e021dc03#diff-cd9d084024219e2261760489591c11f2)(Shout out to @kamilkisiela!) of performing the current logic with rxjs, which might help bundle size.

Second it's a breaking change, since rxjs and zen-observable handle errors thrown inside the [subscriber function](https://github.com/zenparsing/zen-observable#new-observable--subscribe-) differently. rxjs propagates the throw, while zen-observable catches errors and calls `observer.error`. In addition `Observable.map` and `Observable.of` are no longer fields directly on observables, see [here](https://github.com/apollographql/apollo-link/compare/add-rxjs?expand=1#diff-4537b68f0299a56eb7e22e228d1ed32cR248).

The positive is access to a huge amount of functionality and operations of the rxjs ecosystem. My first impression is that rxjs would worthwhile for 2.0 of apollo-link along with some type safety around the context. If you have any use cases, please share!

Thank you @kamilkisiela for the inspiration with [this branch](https://github.com/apollographql/apollo-link/commits/rxjs)!


kamranayub:
This would be great--our use case is that we have Angular services that wrap some business logic services that watch and subscribe to GraphQL queries and notify observers--we have semi-complex rules for how we transform those query responses back to individual observers so using some of the nice RxJS methods to set up a observable chains will be pretty helpful! 👍

kamranayub:
For now I've been using Rx.Observable.from(client.watchQuery) and it's been
working super well.

evans:
@goldenbearkin We'll have to look at the impact on bundle size more closely, since for users outside of the Angular ecosystem the change would mean adding RxJs as a PeerDependency and unfortunately has the reputation of being a large addition. I think it is a safe assumption that most users are using treeshaking, so I'm hopeful that RxJs can make it into 2.0.

The best thing to do in the mean time is follow @kamranayub's great suggestion! 

Is one of you willing to contribute a section in the docs on Observables with note of how to use RxJs? The best place would be in the [concepts overview section](https://www.apollographql.com/docs/link/overview.html). It would be a huge help and I think a better understanding of Observables will help push non-Angular users towards RxJs nicely

kamranayub:
@evans, Yeah, I think I can write something tomorrow.

Also, one thing to note about the approach of wrapping with `Observable.from` is that you will lose the Apollo methods off `ObservableQuery`. It's a known issue with how RxJS behaves at its core and there's not a clean way to guarantee that you will ALWAYS have those methods available (due to RxJS static methods).

However, I have a total hack that totally works and we're using it right now:

```
function getQueryObservableFrom(obs) {
  if (!obs) {
    return null;
  }
  if (obs.ish && obs.ish.queryId) {
    return obs.ish;
  }
  if (!obs.source) {
    return null;
  }
  return getQueryObservableFrom(obs.source);
}

function getVariables(obs) {
  const queryObs = getQueryObservableFrom(obs);

  if (queryObs) {
    return queryObs.variables;
  }
  return {};
}

function refetchQuery(obs, variables) {
    const queryObs = getQueryObservableFrom(obs);

    if (queryObs) {
      return queryObs.refetch(variables);
    }

    return Promise.reject(
      new Error(
        "Could not find underlying Apollo ObservableQuery in Observable stack"
      )
    );
}
```

It takes advantage of the internals of `FromObservable` storing the original `ish` object to the Observable-like instance. This means you can call these helper methods on any Observable that originally came from Apollo.

kamilkisiela:
I did RxJS lifting thing to maintain Apollo specific methods more than a year ago and it's a terrible idea. I did it for apollo-angular. As a main issue this shouldn't be considered is that it won't work correctly with typescript and operators. This is why I changed the approach in apollo-angular 1.0

jbaxleyiii:
@evans can you help me understand why `rxjs` would need to be a peer?

goldenbearkin:
rxjs 6 now shipped with pipeable operators and support tree-shaking which will reduce the output bundle size tremendously.  Will Apollo link consider to use rxjs instead of zen-observable?

kamilkisiela:
@evans I pushed some code today, mostly upgrading to RxJS v6 and some refactoring.

I used Peggy's app to see how much gzipped bundle size have changed:
https://codesandbox.io/s/jvlrl98xw3 (I updated each package to latest version)

before the PR: 183.88 KB
after the PR: 197.54 KB (+13.66 KB)

joshribakoff:
This is a good idea in my opinion. rx has affordances already for retrying, debouncing, throttling, de-duplicating and handling race conditions / in flight requests. You could potentially use these in lieu  of maintaining custom apollo links for these use cases, saving on filesize and developer time. There's also ways to chain rxjs operators that may overlap with operators split, from, and concat which are exported as part of the ApolloLink interface. Its better to use the rxjs terminology instead of making people learn "new" terminology in my opinion.

rxjs seems to also be more widely used and supported than zen observable. By using a more standard observable implementation, it lowers the hurdle for someone new to apollo and could allow for reducing the API surface of apollo link.

 One issue however is backwards compatibility. Technically this should be reserved for a major version correct?

ivasilov-h4:
Hey, this seems like a really good idea. Would you reconsider merging this branch if I update it to latest master?

Issue #433, opened 2018-01-24
Feature Idea (apollo-link-retry):  attempt "fix" before scheduling retries


kiran-sonnad-ck:
In some error cases, there are ways to address the error before attempting a retry. For example, a `401` might be the result of an expired access token that could be refreshed before retrying.

The user could provide a function that would be called prior to scheduling a retry in order to attempt a fix. The fix might be async, so callback or Promises would be required. If the promise resolves (or callback is invoked), scheduling the retry can proceed. Passing the same options that `retryIf` gets would be helpful in case they only want to try on the first retry attempt, or only for certain errors.

pho3nixf1re:
Why not just convert `retryIf` to support a promise? Is there a major technical hurdle involved with that?

pho3nixf1re:
It seems that this section here need to support an optional promise callback to wait on if it is going to retry. Currently all it does is schedule a timeout... Maybe we could provide a custom retry callback instead? That may be the easiest.

https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-retry/src/retryLink.ts#L147-L172

kiran-sonnad-ck:
Adding Promise support to `retryIf` would certainly work, I just wasn't sure whether it might be overloading that function too much (from an API perspective). It's definitely the minimal solution though.

I also hacked together a local version that supports a new `fix` function option that is called before `scheduleRetry`.

pho3nixf1re:
Do you have a branch I can compare with @kiran-sonnad-ck ? I'm working on a fork and PR as a request for comment implementation. I tried implementing my own link for this but it's buggy. We need this pretty quickly so I'm willing to take on the task of seeing this through.

kiran-sonnad-ck:
I'll open a PR that we can work on.

evans:
We merged in #436 and published in 2.2.0 Thank you @ksonnad for adding the ability to use promises in `retryIf`!

PR #435, opened 2018-01-25
feat(apollo-link-retry): handle Promises from `retryIf` and `attempts`

kiran-sonnad-ck:
This enables the consumer to perform async operations when deciding whether to retry, or even to attempt a fix before retrying.

Addresses https://github.com/apollographql/apollo-link/issues/433. This is a WIP while we nail down the API, tests will follow.

FYI @pho3nixf1re

### DIFF
--- a/packages/apollo-link-retry/src/retryFunction.ts
+++ b/packages/apollo-link-retry/src/retryFunction.ts
@@ -5,7 +5,7 @@ import { Operation } from 'apollo-link';
  * response should be retried.
  */
 export interface RetryFunction {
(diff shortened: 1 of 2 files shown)


pho3nixf1re:
Should there be a test for the Promise use case, or is that fairly redundant to the existing test?

pho3nixf1re:
This will definitely handle my use case. Thanks for the quick turnaround on the PR! 🎉

kiran-sonnad-ck:
Closed in favor of #436 (this was opened with the wrong account).

PR #436, opened 2018-01-26
feat(apollo-link-retry): handle Promises from `retryIf` and `attempts`

ksonnad:
This supersedes #435. Sorry for the duplicate PRs, I mistakenly created the other from a work account and wasn't sure whether I could sign the contributor agreement. 

> Handle Promises from `retryIf` and `attempts`.
> 
> This enables the consumer to perform async operations when deciding whether to retry, or even attempt a fix before retrying.

Addresses #433 (CC @pho3nixf1re)

### DIFF
--- a/packages/apollo-link-retry/src/__tests__/retryLink.ts
+++ b/packages/apollo-link-retry/src/__tests__/retryLink.ts
@@ -145,6 +145,7 @@ describe('RetryLink', () => {
     const attemptStub = jest.fn();
     attemptStub.mockReturnValueOnce(true);
     attemptStub.mockReturnValueOnce(true);
+    attemptStub.mockReturnValueOnce(Promise.resolve(true));
     attemptStub.mockReturnValueOnce(false);
 
     const retry = new RetryLink({
@@ -161,6 +162,7 @@ describe('RetryLink', () => {
       [1, operation, standardError],
       [2, operation, standardError],
       [3, operation, standardError],
+      [4, operation, standardError],
     ]);
   });
 });
--- a/packages/apollo-link-retry/src/retryLink.ts
+++ b/packages/apollo-link-retry/src/retryLink.ts
@@ -17,6 +17,13 @@ import {
   buildRetryFunction,
 } from './retryFunction';
 
+function asPromise(result: boolean | Promise<boolean>): Promise<boolean> {
+  if (result instanceof Promise) {
+    return result;
+  }
+  return Promise.resolve(result);
+}
+
 export namespace RetryLink {
   export interface Options {
     /**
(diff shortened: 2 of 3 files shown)


ksonnad:
@nevir and @jbaxleyiii: you appear to be the most recent active contributors. Can I get a general 👍 before proceeding to update the documentation and such?

ksonnad:
@nevir or @jbaxleyiii, how can I move this along? I can't sign the contributor agreement because the page appears to be broken (I get an `500 internal server error` when trying to submit). Is there anything else blocking progress here?

nevir:
Hey, sorry - I think I can take a look at this within the next few days

raarts:
No pressure, but having this would really help my use case too (retry request on 401 due to expired JWT). Unless there's a different way to handle it?

raarts:
@ksonnad, I just signed the contributor agreement, so that page is working again. Thank you to whoever fixed it 👍 .

ksonnad:
Huh, it's still not working for me. I just sent in a PDF to the email address and asked that this PR be unblocked.

ksonnad:
@nevir any progress on this?

raarts:
@ksonnad why is that page not working for you, do you get an error? I go to https://contribute.meteor.com/, I click `sign in with github`, fill in the form and press `submit` and it works. Sending PDF's by email does not seem to work either :-( 
Mentioning @nevir just to be sure.

ksonnad:
Not sure, here's what happens when I try: 
screen shot 2018-02-08 at 6 47 02 am

raarts:
For me it works, and it looks like this. Maybe it's your browser or some plugin?

screen shot 2018-02-08 at 16 14 34

ksonnad:
Chrome, Incognito, Safari, nothing works.

raarts:
Weird, just tried again, still works. Since your case shows a server error, I can't imagine what the cause could be

raarts:
@nevir Bump.

packages/apollo-link-retry/src/__tests__/retryLink.ts
```diff
@@ -145,6 +145,7 @@ describe('RetryLink', () => {
     const attemptStub = jest.fn();
     attemptStub.mockReturnValueOnce(true);
     attemptStub.mockReturnValueOnce(true);
+    attemptStub.mockReturnValueOnce(Promise.resolve(true));
```
nevir: Nit here, but let's break this out into a separate test, so it's easier to focus on if someone breaks logic related to promise support in the future

packages/apollo-link-retry/src/retryLink.ts
```diff
@@ -17,6 +17,13 @@ import {
   buildRetryFunction,
 } from './retryFunction';
 
+function asPromise(result: boolean | Promise<boolean>): Promise<boolean> {
+  if (result instanceof Promise) {
```
nevir: There are still enough folks out there using alternative `Promise` implementations that this may end up being dangerous.  I'd recommend duck typing here…

```ts
if ('then' in result && typeof result.then === 'function') {
```

…or letting TypeScript juggle it for you by converting `onError` to an `async` function, and awaiting the result of `retryIf` (you can `await` static values or promises)

nevir:
cc @jbaxleyiii - who's the right person to help out here w/ the CLA bot?

ksonnad:
@nevir good points, I've addressed your comments.

nevir:
Perfect!  Last thing you'll need to do is bump up the [expected bundle size](https://github.com/apollographql/apollo-link/blob/master/package.json#L63) (let's go with 2.5Kb), sync with master, and then we're good 👍

ksonnad:
Done! Thanks for moving this along.

Once merged, how long might a release take?

evans:
@ksonnad published in 2.2.0. I added the changes to the changelog, so no worries on that front. Can you document the change [here](https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-retry/README.md)? 

@nevir Thank you for seeing this through!

PR #445, opened 2018-01-29
docs(apollo-link-state): update to fix warnings

wKovacs64:
Just bringing over a documentation update per @peggyrayzis:

https://github.com/apollographql/apollo-link-state/pull/177#issuecomment-359936781


peggyrayzis:
Thanks so much @wKovacs64!

PR #458, opened 2018-02-02
fix wrong link to documentation

kadmelia:
Readme file has a link to https://www.apollographql.com/docs/link/docs/link/index.html which seems to not exists, right link seems to be https://www.apollographql.com/docs/link/index.html.


evans:
I go to first #459. Thank you so much for making the fix too!

PR #463, opened 2018-02-05
Expose execute on ApolloLink as static

imranolas:
TODO:

## Why?
Apollo link is super convenient tooling for composing any request/response cycle for GraphQL. However, with `#execute` being it's own import it forces other libraries to take on the dependency responsibility if they want to support Link. This way we can allow users to take responsibility for the dep and simply call execute on the exposed public method whenever a link is used.

### DIFF
--- a/packages/apollo-link/CHANGELOG.md
+++ b/packages/apollo-link/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### vNEXT
 
+- Expose `#execute` on ApolloLink as static
+
 # 1.0.7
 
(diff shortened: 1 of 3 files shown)


jbaxleyiii:
@imranolas this is a great idea! I'm all for this!

jbaxleyiii:
@imranolas would you mind adding a smoke test by changing one of the `executes` somewhere in the test suite to be from the static instead of the import?

imranolas:
Sure can :)

PR #468, opened 2018-02-07
Error http mixed data

jbaxleyiii:
This fixes a bug reported by @wesbos where a 400 (or network error) wasn't sending data to the component that also came back from the GraphQL endpoint. Now data + errors will fire if there is a parse able result, then trigger the network error event.


wesbos:
Oh hey - glad to hear this :D

Issue #389, opened 2018-01-05
zen-observable breaks while compiling.


VrajSolanki:
**Intended outcome:**

I updated the apollo-client and added apollo-link to my React-Native app. But the apollo-link breaks during compiling phase, but I expected the app to run smoothly as I have already updated the same in another react app.

react-native version : 0.49.1
apollo-link version: 1.0.7

**Code for apollo Client**

```
import { ApolloClient } from 'apollo-client'
import { HttpLink, createHttpLink } from 'apollo-link-http';
import { setContext } from 'apollo-link-context';
import { InMemoryCache, IntrospectionFragmentMatcher } from 'apollo-cache-inmemory';
import schema from './schema.json'

const httpLink = createHttpLink({ uri: `localhost:3000/graphql` });

const cache = new InMemoryCache({
  fragmentMatcher: new IntrospectionFragmentMatcher({
    introspectionQueryResultData: schema.data
  }),
});

const client = new ApolloClient({
  link: httpLink,
  cache: cache,
});

export default client

```
**Actual outcome:**
![screen shot 2018-01-05 at 8 32 41 pm](https://user-images.githubusercontent.com/28093589/34614833-c666dfa0-f258-11e7-8e21-3867bdac0d1b.png)

```
error: bundling failed: TypeError: pathToNodeModules/node_modules/zen-observable/zen-observable.js: Property left of AssignmentExpression expected node to be of a type ["LVal"] but instead got "ConditionalExpression"
    at Object.validate (pathToNodeModules/node_modules/babel-types/lib/definitions/index.js:109:13)
    at Object.validate (pathToNodeModules/node_modules/babel-types/lib/index.js:505:9)
    at NodePath._replaceWith (pathToNodeModules/node_modules/babel-traverse/lib/path/replacement.js:176:7)
    at NodePath.replaceWith (pathToNodeModules/node_modules/babel-traverse/lib/path/replacement.js:160:8)
    at PluginPass.MemberExpression (pathToNodeModules/node_modules/babel-preset-react-native/transforms/transform-symbol-member.js:33:14)
    at newFn (pathToNodeModules/node_modules/babel-traverse/lib/visitors.js:276:21)
    at NodePath._call (pathToNodeModules/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (pathToNodeModules/node_modules/babel-traverse/lib/path/context.js:48:17)
[...]

PR #489, opened 2018-02-16
fix rollup imports and sourcemap support(jest + rollup)

### LINKED ISSUE #248: zen-observable import breaks queries in apollo-client when transpiled
Hi there!

I'm working to get a new project set up with Apollo Client, Apollo Link, babel, and webpack. I use a pretty simple project template to set it up with the documented suggestions in Apollo Client. When I go to run it and send it a query, I get a runtime exception and queries cannot go through. I traced this to how the `zen-observable` package is imported.

**Intended outcome:**

Apollo Client can make a query using Apollo Link, using the following code:

```javascript
import { ApolloClient } from "apollo-client"
import { HttpLink } from "apollo-link-http"
import { InMemoryCache } from "apollo-cache-inmemory"
import { gql } from "graphql-tag"

const graphqlClient = new ApolloClient({
        link: new HttpLink(),
        cache: new InMemoryCache(),
})
graphqlClient.query({ query: gql`query { counter { count } }` })
        .then(console.log)
        .catch(console.error)
```

**Actual outcome:**

Apollo Client throws the following exception:

```
TypeError: _super.call is not a function
Stack trace:
ObservableQuery@webpack-internal:///16:56:21
QueryManager</QueryManager.prototype.watchQuery@webpack-internal:///94:404:16
QueryManager</QueryManager.prototype.query/resPromise<@webpack-internal:///94:431:20
QueryManager</QueryManager.prototype.query@webpack-internal:///94:429:26
ApolloClient</ApolloClient.prototype.query@webpack-internal:///93:102:16
@webpack-internal:///59:14:1
[59]@http://host/assets/client.js:16:1
__webpack_require__@http://host/assets/manifest.js:713:12
fn@http://host/assets/manifest.js:118:20
[49]@http://host/assets/client.js:7:18
__webpack_require__@http://host/assets/manifest.js:713:12
webpackJsonpCallback@http://host/assets/manifest.js:26:23
@http://host/assets/client.js:1:1
```

The `ObservableQuery` imports the `Observable` from the Apollo Link project, which imports it from the `zen-observable` package. When transpiled to JavaScript, the apollo-link package has this at `lib/index.js`:

```javascript
import * as Observable from 'zen-observable';
export { Observable };
```

It seems that this causes some strange case when transpiling and later subclassing this Observable, but this seems to be where the `call` method stops being available through my investigation.

Attached is a test project which should demonstrate the issue. I assume Node.js v6 or v8 installed with Yarn. [Test Project](https://github.com/apollographql/apollo-link/files/1473237/apollo-link-test.zip)

1. Unzip and cd into the test project
2. `yarn install`
3. `yarn run start` and open http://localhost:3000/
4. You should see the error shown above in the console.

You can workaround this by changing `node_modules/apollo-link/lib/index.js:3` from:

`import * as Observable from 'zen-observable';`

to:

`import Observable from 'zen-observable';`

The problem is if you make this change in the corresponding .ts file, it causes a linting failure, and I'm not sure how best to resolve that.

There is [this issue on StackOverflow](https://stackoverflow.com/questions/47156532/apollo-client-2-with-react-couldnt-make-query) which suggests that other people are having this issue.

Thank you so much for your time, for looking at my bug report, and for the awesome work on Apollo!

jbaxleyiii:
Fixes #248 

cc @evans

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "lint-staged": "6.1.1",
     "pre-commit": "1.2.2",
     "prettier": "1.10.2",
+    "rollup-plugin-sourcemaps": "^0.4.2",
     "ts-jest": "21.2.4",
     "typescript": "2.7.2"
   }
--- a/packages/apollo-link-batch-http/package.json
+++ b/packages/apollo-link-batch-http/package.json
@@ -71,6 +71,7 @@
     },
     "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|tsx|js)$",
     "moduleFileExtensions": ["ts", "tsx", "js", "json"],
-    "testPathIgnorePatterns": ["/node_modules/", "sharedHttpTests.ts"]
+    "testPathIgnorePatterns": ["/node_modules/", "sharedHttpTests.ts"],
+    "mapCoverage": true
   }
 }
--- a/packages/apollo-link-batch-http/src/__tests__/batchHttpLink.ts
+++ b/packages/apollo-link-batch-http/src/__tests__/batchHttpLink.ts
@@ -1,6 +1,6 @@
 import { ApolloLink, execute, Observable, makePromise } from 'apollo-link';
 import { print } from 'graphql';
-import * as fetchMock from 'fetch-mock';
+import fetchMock from 'fetch-mock';
 import gql from 'graphql-tag';
 
 import { sharedHttpTest } from './sharedHttpTests';
--- a/packages/apollo-link-batch/package.json
+++ b/packages/apollo-link-batch/package.json
@@ -65,6 +65,7 @@
       ".(ts|tsx)": "<rootDir>/node_modules/ts-jest/preprocessor.js"
     },
     "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|tsx|js)$",
-    "moduleFileExtensions": ["ts", "tsx", "js", "json"]
+    "moduleFileExtensions": ["ts", "tsx", "js", "json"],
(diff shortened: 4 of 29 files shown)


evans:
@jbaxleyiii I added sourcemap support for [jest coverage](https://github.com/kulshekhar/ts-jest#coverage) and rollup builds(wasn't picking up the tsc sourcemaps). Tested with example from #248, so should work :pray:. We should definitely think about moving to RxJs if they have addressed the bundle size issues

evans:
apollo-link's bundle now emits `Observable = Observable && Observable.hasOwnProperty('default') ? Observable['default'] : Observable;` and the index contains `import Observable from 'zen-observable';
`, so we should be set

littletower:
awesome! when will this be merged? :blush: thanks!

jbaxleyiii:
@evans AMAZING

evans:
@littletower We merged and published. Can you post in #389 if it still doesn't work with 1.2.0

JounQin:
It breaks https://github.com/apollographql/apollo-link/issues/511, and requires 
adding `allowSyntheticDefaultImports: true` into `tsconfig.json`, is that reasonable to require specific tsconfig?

evans:
@JounQin Good point! We shouldn't be forcing a specific `tsconfig.json`. I think the best way forward is to either switch to RxJs or write an implementation of zen-observable in typescript, since the `* as` imports break js builds and a direct import necessitates the `allowSyntheticDefaultImports`. Please take a look at #515. If you think there is a better way forward please share.

PR #504, opened 2018-02-18
Update batch http link example in README

matteodem:
To latest version with class called BatchHttpLink. Small change in the README file


evans:
Good catch! Thank you @matteodem

PR #515, opened 2018-02-21
zen-observable implementation in Typescript to remove import issues

### LINKED ISSUE #389: zen-observable breaks while compiling.

### LINKED ISSUE #476: "TypeError: forward(...).flatMap is not a function" with apollo-link 1.1.0.
Hey guys!

We've got a custom middleware that takes a mutation operation and keeps it in a queue for an undetermined amount of time. The use-case is a CQRS-style task-based command API where confirmations are received asynchronously through a websocket, instead of "synchronously" as mutation response. The code looks something like this:

```js
return forward(operation).flatMap(response =>
  new Observable(observer => {
    putInQueueUntilWeReceiveConfirmation(response).then(observer.next).then(observer.complete);
  })
);
```

Up until `1.0.7` this worked fine, however, starting from `1.1.0` this is throwing a `TypeError: forward(...).flatMap is not a function` exception. Did anything change with the way operation observables are passed? I can't seem to find anything in the changelog.

Downgrading `apollo-link` to `1.0.7` fixes this issue, so it is definitely introduced in `1.1.0`.

### LINKED ISSUE #511: Error when building with v1.2.0: "@types/zen-observable/index has no default export"
**Intended outcome:**

Build Angular CLI project with apollo-link. Looks like the latest release has issues with types/zen-observable

**Actual outcome:**

Got this error: 

```bash
ERROR in node_modules/apollo-link-context/node_modules/apollo-link/lib/types.d.ts(2,8): error
TS1192: Module '"/tmp/build_6e1e79a9d4e4cc0fceeda16a93512a8c/node_modules/@types/zen
observable/index"' has no default export.
```

Typical Angular CLI setup with apollo-client v2.2.5 and apollo-angular v1.0.1. Only happens with apollo-link v1.2.0

evans:
We need an implementation of zen-observable in typescript, since importing the current `zen-observable` library breaks when this library and apollo-client are imported. `* as Observable` imports break js builds(#248) and a direct import necessitates the `allowSyntheticDefaultImports` in users `tsconfig.json`s(#511). [This comment](https://github.com/apollographql/apollo-link/pull/457#issuecomment-364169732) suggests we could use `import Observable = require('zen-observable')`, however it requires emitting as a non-ES module, such as CJS.

This PR contains a typescript wrapper around a `require`d zen-observable, since the wrapper allows us to import zen observable in an interoperable way and gain the typings.

The ultimate goal is to move to RxJs(#380) due to its stability and the extensiveness of its operators. The main concern was the bundle size. With Lettable operators this is no longer a concern.

Fixes #476 Fixes #511 Fixes #389

### DIFF
--- a/packages/zen-observable-ts/src/zenObservable.ts
+++ b/packages/zen-observable-ts/src/zenObservable.ts
@@ -1,5 +1,12 @@
 declare function require(name: string);
-const _Observable = require('zen-observable');
+
+const Observable: {
+  new <T>(subscriber: Subscriber<T>): Observable<T>;
+  from<R>(...args: Array<R>): Observable<R>;
+  of<R>(...args: Array<R>): Observable<R>;
+} = require('zen-observable');
+
+export default Observable;
 
 import { ZenObservable } from './types';
 
@@ -9,63 +16,25 @@ export type Observer<T> = ZenObservable.Observer<T>;
 export type Subscriber<T> = ZenObservable.Subscriber<T>;
 export type ObservableLike<T> = ZenObservable.ObservableLike<T>;
 
-export default class Observable<T> extends (_Observable as {
-  new (subscriber: any): any;
-  from(...args): any;
-  of(...args): any;
-}) {
-  public static from<R>(
-    observable: Observable<R> | ZenObservable.ObservableLike<R> | ArrayLike<R>,
-  ): Observable<R> {
-    return super.from(observable);
-  }
-
-  public static of<R>(...items: R[]): Observable<R> {
-    return super.of(...items);
-  }
-
-  constructor(subscriber: ZenObservable.Subscriber<T>) {
-    super(subscriber) as any;
-  }
-
-  public subscribe(
+export interface Observable<T> {
+  subscribe(
     observerOrNext: ((value: T) => void) | ZenObservable.Observer<T>,
     error?: (error: any) => void,
     complete?: () => void,
-  ): ZenObservable.Subscription {
-    return super.subscribe(observerOrNext, error, complete);
-  }
+  ): ZenObservable.Subscription;
 
-  public forEach(fn: (value: T) => void): Promise<void> {
-    return super.forEach(fn);
-  }
+  forEach(fn: (value: T) => void): Promise<void>;
 
-  public map<R>(fn: (value: T) => R): Observable<R> {
-    return super.map(fn);
-  }
+  map<R>(fn: (value: T) => R): Observable<R>;
 
-  public filter(fn: (value: T) => boolean): Observable<T> {
-    return super.filter(fn);
-  }
+  filter(fn: (value: T) => boolean): Observable<T>;
 
-  public reduce<R = T>(
+  reduce<R = T>(
     fn: (previousValue: R | T, currentValue: T) => R | T,
     initialValue?: R | T,
-  ): Observable<R | T> {
-    return super.reduce(fn, initialValue);
-  }
+  ): Observable<R | T>;
 
-  public concat(...sources: Array<Observable<T>>) {
-    return super.concat(...sources);
-  }
+  concat(...sources: Array<Observable<T>>);
 
-  public flatMap<R>(
-    fn: (value: T) => ZenObservable.ObservableLike<R>,
-  ): Observable<R> {
-    if (this.observable.flatMap) {
-      return super.flatMap(fn);
-    } else {
-      console.error('this zen-observable does does not support flatMap');
-    }
-  }
+  flatMap<R>(fn: (value: T) => ZenObservable.ObservableLike<R>): Observable<R>;
 }
--- a/packages/zen-observable-ts/src/zenObservable.ts
+++ b/packages/zen-observable-ts/src/zenObservable.ts
@@ -1,0 +1,528 @@
+import { ZenObservable } from './types';
+
+export { ZenObservable };
+
+export type Observer<T> = ZenObservable.Observer<T>;
+export type Subscriber<T> = ZenObservable.Subscriber<T>;
+export type ObservableLike<T> = ZenObservable.ObservableLike<T>;
+
+// === Abstract Operations ===
+function cleanupSubscription(subscription: Subscription) {
+  // Assert:  observer._observer is undefined
+
+  let cleanup = subscription._cleanup;
+
+  if (!cleanup) {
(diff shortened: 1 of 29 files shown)


evans:
This solution is definitely non-ideal, since it requires maintaining another implementation of zen-observable. If there is a better way forward, please let me know

evans:
@benjamn pointed out that we can import `zen-observable` with a `require` interoperable, so since we already have the types written, we area able to depend on the logic of the zen-observable package. 

This solution seems like the best way to go, since it removes the need to reimplement and we have typing information.

I still have questions about interoperability with other zen-observable's, such as will `instanceof` and symbols still work.

packages/zen-observable-ts/src/zenObservable.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import { ZenObservable } from './types';
+
+export { ZenObservable };
+
+export type Observer<T> = ZenObservable.Observer<T>;
+export type Subscriber<T> = ZenObservable.Subscriber<T>;
+export type ObservableLike<T> = ZenObservable.ObservableLike<T>;
+
+export default class Observable<T> {
```
benjamn: Since you mentioned `instanceof`, it might be possible to make this class inherit from `_Observable`:
```typescript
export default class Observable<t> extends _Observable {
```

packages/zen-observable-ts/src/zenObservable.ts
```diff
@@ -0,0 +1,69 @@
...
+  public subscribe(
+    observerOrNext: ((value: T) => void) | ZenObservable.Observer<T>,
+    error?: (error: any) => void,
+    complete?: () => void,
+  ): ZenObservable.Subscription {
+    return this.observable.subscribe(observerOrNext, error, complete);
+  }
+
+  public forEach(fn: (value: T) => void): Promise<void> {
+    return this.observable.forEach(fn);
```
benjamn: If `Observable<T>` extends `_Observable`, then methods implementations like this would become
```typescript
return super.forEach(fn);
```
because `this instanceof _Observable`.

packages/zen-observable-ts/src/zenObservable.ts
```diff
@@ -0,0 +1,69 @@
...
+  ): Observable<R> {
+    return _Observable.from(observable);
+  }
+
+  public static of<R>(...items: R[]): Observable<R> {
+    return _Observable.of(...items);
+  }
+
+  constructor(subscriber: ZenObservable.Subscriber<T>) {
+    this.observable = new _Observable(subscriber);
```
benjamn: And here you would just call `super(subscriber)`.

packages/zen-observable-ts/src/zenObservable.ts
evans: Looks like `extend _Observable` causes [ts] 'extends' clause of exported class 'Observable' has or is using private name '_Observable'. I'm not overly worried about this change, since it is an implementation detail that leaks out and `instanceof` will still  work within the `apollo-link`'s exported observable

PR #522, opened 2018-02-26
Add missing global Rollup alias for apollo-link-http-common

### LINKED ISSUE #521: No global alias defined for apollo-link-http-common in rollup.config.js
**Intended outcome:**

When using `apollo-link` bundles built for the browser, `apollo-link` should be able to access utilities such as `checkFetcher` from `apollo-link-http-common`.

**Actual outcome:**

As you can see in the [bundle.umd.js](https://unpkg.com/apollo-link-http@1.5.1/lib/bundle.umd.js), `apollo-link-http` expects to find the common HTTP link utilities in `global.apolloLinkHttpCommon`. 

However, `apollo-link-http-common` is configured to export everything via `global.apolloLink.utilities` (see its [Rollup config](https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-http-common/rollup.config.js#L3)).

The reason for this is that there is no alias defined in the top-level [rollup.config.js](https://github.com/apollographql/apollo-link/blob/master/rollup.config.js) for `apollo-link-http-common`. When building `apollo-link-http` with `yarn build` it even warns about this:

```sh
➜  apollo-link-http git:(master) yarn build
yarn run v1.3.2
$ npm run clean

> apollo-link-http@1.5.1 clean /Users/jannis/Work/oss/apollographql/apollo-link/packages/apollo-link-http
> rimraf lib/* && rimraf coverage/*

$ tsc -p .
$ npm run bundle

> apollo-link-http@1.5.1 bundle /Users/jannis/Work/oss/apollographql/apollo-link/packages/apollo-link-http
> rollup -c


lib/index.js → lib/bundle.umd.js...
No name was provided for external module 'apollo-link-http-common' in options.globals – guessing 'apolloLinkHttpCommon'
```

The build warning can be reproduced simply by building `apollo-link-http` from the repository source tree.

The runtime issue with `global.apolloLinkHttpCommon` not being found can be reproduced by loading the bundles of both `apollo-link-http-common` and `apollo-link-http` in the browser and running `apolloLink.http.createHttpLink(...)`—it will throw the following error:

```Uncaught TypeError: Cannot read property 'checkFetcher' of undefined```

Jannis:
This fixes #521, making it so that the apollo-link-http bundle resolves
apollo-link-http-common to global.apolloLink.utilities instead of
global.apolloLinkHttpCommon (which does not exist).

### DIFF
--- a/packages/apollo-link-http-common/rollup.config.js
+++ b/packages/apollo-link-http-common/rollup.config.js
@@ -1,3 +1,3 @@
 import build from '../../rollup.config';
 
-export default build('utilities');
+export default build('httpCommon');
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,6 +6,7 @@ export const globals = {
   'apollo-link': 'apolloLink.core',
   'apollo-link-batch': 'apolloLink.batch',
   'apollo-utilities': 'apollo.utilities',
+  'apollo-link-http-common': 'apolloLink.utilities',
   'zen-observable-ts': 'apolloLink.zenObservable',
 
   //GraphQL
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,7 +6,7 @@ export const globals = {
   'apollo-link': 'apolloLink.core',
   'apollo-link-batch': 'apolloLink.batch',
   'apollo-utilities': 'apollo.utilities',
-  'apollo-link-http-common': 'apolloLink.utilities',
+  'apollo-link-http-common': 'apolloLink.httpCommon',
   'zen-observable-ts': 'apolloLink.zenObservable',
 
   //GraphQL


evans:
@Jannis Thank you for flagging the issue and the PR! I think we should change the name of `apollo-link-http-common`'s rollup target to `apolloLink.httpCommon`. Could you change `utilities` to `httpCommon` [here](https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-http-common/rollup.config.js#L3) and update the global rollup config?

Hopefully that will allow Netlify to redeploy correctly too 😄

Jannis:
@evans Sure, sounds like a good idea!

Jannis:
@evans Done. It's two separate commits right now but I'm happy to squash them together if that's what you prefer.

evans:
@Jannis We've use separate commits for the most part in the past, so this works great. Thank you so much!

PR #555, opened 2018-03-22
doc: fix composition doc (useless, obsolete Retry)

fredericgermain:
Hi !

Updating this bit of documentation, it seems obsolete.

Another example which would be useful is the one in the client doc for subscription -> https://www.apollographql.com/docs/react/features/subscriptions.html


evans:
@fredericgermain Thank you for opening the pull request! I'm curious what gave you the impression that this pattern is obsolete, since it should work as a common way of having retrying two different http links. Does it need a diagram?

Including the subscriptions example as well would be great! Could you make that change in this PR or another?

fredericgermain:
There is no more `Retry` class is exported from `apollo-link-retry` package. It was renamed to `RetryLink`

cf
* https://github.com/apollographql/apollo-link/tree/master/packages/apollo-link-retry
* https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-retry/src/retryLink.ts

At the Retry/RetryLink would not bring anything useful here, it's better to have a simpler example in my opinion.

Or maybe I'm missing something?

PR #584, opened 2018-03-29
Http: Check for signal on fetch options before adding own abort controller

evans:
Currently if the user passes in a signal option, it gets overwritten if a controller is successfully made. This should not be the case as something outside of the http links should be able to abort the fetch request. 

**Pull Request Labels**

/label bug

### DIFF
--- a/packages/apollo-link-batch-http/src/batchHttpLink.ts
+++ b/packages/apollo-link-batch-http/src/batchHttpLink.ts
@@ -121,8 +121,12 @@ export class BatchHttpLink extends ApolloLink {
         return fromError<FetchResult[]>(parseError);
       }
 
-      const { controller, signal } = createSignalIfSupported();
-      if (controller) (options as any).signal = signal;
+      let controller;
+      if (!(options as any).signal) {
+        const { controller: _controller, signal } = createSignalIfSupported();
+        controller = _controller;
+        if (controller) (options as any).signal = signal;
+      }
 
       return new Observable<FetchResult[]>(observer => {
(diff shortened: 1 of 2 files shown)


helfer:
@evans I'm about a year too late and I may not have all the context, but shouldn't clients who want to abort a request do so by unsubscribing from the observable in an orderly fashion? Reaching into the fetch and cancelling the request from outside breaks the link abstraction and should be strongly discouraged, if not made impossible. For example, it's not super clear to me what happens if you have a retry link in front of the HTTP link, and then reach in and abort the fetch? The retry link will keep retrying, and presumably the fetches will continue to fail due to the abort signal, correct?
Another problem is if you have a link that somehow multiplexes requests (dedupe, debounce, etc.). In that case, other requests may have been multiplexed to the same fetch. If that original requests isn't interested any more and aborts the fetch, all requests will fail, which isn't what you would expect.
It seems to me like we should really discourage people from reaching into the http link and aborting a fetch via the signal, and instead just unsubscribe in an orderly fashion, which should abort the request as well.

ChristianJacobsen:
@helfer Sorry for reviving a dead thread, but how would you go about doing what you propose? In my scenario I've replaced fetch with axios using [buildAxiosFetch](https://github.com/lifeomic/axios-fetch) so that I can get the progress of a file upload. Even when I pass in a signal to the `fetchOptions` of the mutation I'm not able to cancel the request. Afaik the only time I've interacted with observables when using Apollo is when I've been using Subscriptions, and then it's as easy as calling `.unsubscribe()`. How would one go about doing this for a mutation?

PR #596, opened 2018-04-09
Fix cache.writeData call in apollo-link-state doc

aanand:
As I understand `cache.writeData`, it expects an object with a `data` property as its first argument. There's a code example in `state.md` that passes in the data object unwrapped, which led me astray.

(I take it this doesn't require a `CHANGELOG` addition anywhere, being a docs change?)

- [ ] feature
- [ ] blocking
- [x] docs


evans:
@aanand good catch! Thank you

Issue #542, opened 2018-03-10
ApolloLink emits both value and error in catch


bmsantos:
**Expected Behavior**

Observable.catch should emit either recovered value or error.

**Actual Behavior**

The error is emitted and bubbled up the Rx Observable chain just fine. When caught by the subscriber error action, the observable will be closed and the value emition is never handled causing the app to crash.

**A _simple_ reproduction**

Place any query and ensure a response with HTTP status code set to anything above 300.
Make sure the errors is set and not null.

**Issue Labels**

- [ ] has-reproduction
- [ ] feature
- [X] blocking
- [X] good first issue

Some details:

I believe the issue is in: https://github.com/apollographql/apollo-link/blob/c9843e5a051acbc5e1760d5f9a800bdc369f1e8f/packages/apollo-link-http/src/httpLink.ts#L130

```typescript
        ...
        .catch(err => {
          if (err.name === 'AbortError') return;
          if (err.result && err.result.errors && err.result.data) {
            observer.next(err.result);
          } 
          observer.error(err);
        });
```

The above code when modified to the sample below fixes the experienced issues:

```typescript
        ...
        .catch(err => {
          if (err.name === 'AbortError') return;
          if (err.result && err.result.errors && err.result.data) { // Recover from error and emit value
            observer.next(err.result);
          }  else { // Or just emit error and close stream
            observer.error(err);
          }
        });
```

melloflavio:
I've submitted a [PR#606](https://github.com/apollographql/apollo-link/pull/606) for this fix. What else is needed to incorporate this to `develop`

PR #606, opened 2018-04-18
Fix double error thrown if http_status > 300 && result has data

melloflavio:
Issue is described in detail in [#542](https://github.com/apollographql/apollo-link/issues/542).

In summary:
If `apollo-link-http` receives an error but the response has data to be presented, it will call `.next(err.result);` in the request observable and then call `.error()` in the same observable.
Since the 1st call (`.next(...)`) ends up closing the observable/stream, the 2nd call (`.error()`) will throw an error which is pretty much uncatchable. This is exceptionally bad when using Apollo in ReactNative since this exception crashes the application.

The solution simply `return` after the call to `.next(...)` to avoid the late call to `.error()`.

TODO:

**Pull Request Labels**

- [ ] feature
- [x] blocking
- [ ] docs

### DIFF
--- a/packages/apollo-link-http/src/httpLink.ts
+++ b/packages/apollo-link-http/src/httpLink.ts
@@ -159,6 +159,7 @@ export const createHttpLink = (linkOptions: HttpLink.Options = {}) => {
             // in the UI you want to show data where you can, errors as data where you can
             // and use correct http status codes
             observer.next(err.result);
+            return;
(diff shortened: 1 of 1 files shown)


melloflavio:
I'm not sure how would I go about writing a test for this specific bug, given that the symptom is an exception which we are unable to properly catch. 
I accept suggestions...

PR #618, opened 2018-04-27
docs: Add apollo-link-computed-property to community links

czystyl:
# Summary 

Add `apollo-link-computed-property` to community links.

- [x] docs


evans:
@czystyl Thank you! So excited to see how this develops

Issue #558, opened 2018-03-23
zen-observable-ts combines 'export' and 'require'


acoreyj:
**Expected Behavior**
Can package zen-observable-ts with Rollup

**Actual Behavior**
zen-observable-ts results in 
` ReferenceError: require is not defined`

This is because the compiled zenObservable.js is

`export var Observable = require('zen-observable');`

As you can see here https://github.com/rollup/rollup/issues/1058#issuecomment-254187433 you shouldn't combine export and require in the same file as commonjs will ignore the file if it has export/import.

It should compile to something like 

```
import * as zenObservable from 'zen-observable';
export var Observable = zenObservable;
```

acoreyj:
added PR that fixes the issue

evans:
@acoreyj Thanks for opening the PR and Issue! I'm not sure how to rectify this, since using the import syntax will break React Native. See #389, #248, and #515. Do you have an example project that I can clone?

acoreyj:
@evans 
Yup I did just run into #248  with my fix. I updated my PR to what I think is an actual fix. Basically we need to export the default of zen-observable. 

I created a repo showing this error here https://github.com/acoreyj/apollo-link-observable-rollup-issue

See the readme for more details.

Definitely some strange behavior here, my guess is stemming from how zen-observable handles it's exports? Possibly fixed in most recent version?

gaberudy:
This is definitely an issue and I am experiencing it myself with the rollup builds.

This is very painful, hopefully this PR can be merged in.

acoreyj:
in the (https://github.com/apollographql/apollo-link/issues/558) Pull Request I added the allowSyntheticDefaultImports to the tsConfig so now the fixed zenObservable.js will look more normal while doing the same thing.

Note I haven't tested this change with react-native but it definitely fixes #248

```
import zenObservable from 'zen-observable';
export var Observable = zenObservable;
```

bennypowers:
Anything I can do to help this along?

acoreyj:
@bennypowers 

Not sure what else can be done, repo to reproduce is [here](https://github.com/acoreyj/apollo-link-observable-rollup-issue) and PR to fix is [here](https://github.com/apollographql/apollo-link/pull/559)

In the past they mentioned switching to rxjs which would resolve this problem, perhaps that is being actively worked on.

evans:
[...]

PR #623, opened 2018-05-05
Fix zen-observable typescript import

dobesv:
Because zen observable uses the `export = Observable`, it should be imported using the form  `import Observable = require('zen-observable');`.

See https://www.typescriptlang.org/docs/handbook/modules.html under the heading "export = and import = require()":

> When exporting a module using `export =`, TypeScript-specific `import module = require("module")` must be used to import the module.

This error occurs if your tsconfig has 

```
    "esModuleInterop": true,
```

Setting that option to false prevents the error.

### DIFF
--- a/packages/apollo-link/src/index.ts
+++ b/packages/apollo-link/src/index.ts
@@ -8,5 +8,5 @@ export {
 } from './linkUtils';
 export * from './types';
 
-import Observable from 'zen-observable-ts';
+import Observable = require('zen-observable');
 export { Observable };


dobesv:
See also https://github.com/apollographql/apollo-link/issues/558

dobesv:
See also https://github.com/apollographql/apollo-link/issues/248

dobesv:
See also https://github.com/apollographql/apollo-link/pull/380 as a possible alternative fix.

dobesv:
See also https://github.com/apollographql/apollo-link/pull/515

dobesv:
See also https://github.com/apollographql/apollo-link/issues/511

THoisington:
I'm not sure if it is appropriate to comment here, but I have a suspicion that this fix could solve an error that has haunted me for the last three days: 

`ERROR in [at-loader] ./node_modules/@types/zen-observable/index.d.ts:8:9 
    TS2687: All declarations of 'observable' must have identical modifiers.`

Really hope this PR can be merged.

dobesv:
It seems like recent changes to master have probably made this PR obsolete.

nicky-lenaers:
Same issue as @THoisington , still present in 1.2.2

dobesv:
@nicky-lenaers You should probably comment on one of the related issues.  This PR would have to be redone since the code it was based on has changed in a significant way.

THoisington:
@nicky-lenaers There is also a [PR in zen-observable](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26239#issuecomment-395925181). For me it was specifically @types/node adding readonly and @types/zen-observable was missing it. Seems like things can really get messy with @types!

PR #634, opened 2018-05-23
chore: update @types/graphql to 0.13.1

### LINKED ISSUE #601: Two different types with this name exist errors due to mismatch @types/graphql
We getting following errors, it could be due to `@types\graphql` happens to be in `dependencies` rather than in `devDependencies`.
https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link/package.json#L41

This forcing us to use a version that comes with `apollo-link` only. Wondering whether anyone caught this thing yet.

```

graphql-schema/src/__tests__/integration_setup.ts(23,50)
      TS2345: Argument of type 'DocumentNode' is not assignable to parameter of type 'DocumentNode'.
  Types of property 'loc' are incompatible.
    Type 'Location' is not assignable to type 'Location'. Two different types with this name exist, but they are unrelated.
      Types of property 'startToken' are incompatible.
        Type 'Token' is not assignable to type 'Token'. Two different types with this name exist, but they are unrelated.
          Types of property 'kind' are incompatible.
            Type 'TokenKind' is not assignable to type '"<SOF>" | "<EOF>" | "!" | "$" | "(" | ")" | "..." | ":" | "=" | "@" | "["| "]" | "{" | "|" | "}"...'.
              Type '"BlockString"' is not assignable to type '"<SOF>" | "<EOF>" | "!" | "$" | "(" | ")" | "..." | ":" | "=" | "@"| "[" | "]" | "{" | "|" | "}"...'.
```

victorandree:
This is dependency for `apollo-link` which means users of that package cannot use @types/graphql@0.13.1 unless it is updated here (or you get "unrelated" type errors).

Fixes #601 for now. Like `graphql` itself, the `@types/graphql` package should _maybe_ actually be a `peerDependency`.

/label chore


abhiaiyer91:
Thanks for this PR!

amacleay:
This will solve an issue for us in my project, but looks like the PR may be stalled... any way I can help?

victorandree:
Have yet to hear from project maintainers about this... The more I think about it, the more it seems like `@types/graphql` should be a `peerDependency` as long as `graphql` itself is.

amacleay:
If `@types/graphql` is a peer dependency, I believe that would force non-TS consumers to use it, which isn't appropriate.

It shouldn't be necessary to specify as a peer, because if the consumer has `graphql`, they'll need to add `@types/graphql` for their project to compile.  I believe the dependency should just be removed (and moved to `devDependencies`

victorandree:
@amacleay Yeah, on second thought, I would agree with you. This package should not expose `@types/graphql` as a dependency at all, since it doesn't re-export anything from `graphql` (if it did, `graphql` should be a production dependency and *then* `@types/graphql` should be a production dependency).

I would be inclined to update the PR along this reasoning but would like to hear from maintainers regardless.

amacleay:
Very good. I also opened https://github.com/apollographql/apollo-link/pull/641 which just the minimal thing: moving `@types/graphq`l to devDependencies, in case that is more palatable. FWIW, `apollo-link` is the only package in the repo which has `@types/graphql` as a dependency, but many have it as a devDependency.

victorandree:
Updated to merge latest master.

@jbaxleyiii It would be very good for us if this or #641 were merged as we cannot update our @types/graphql without it.

victorandree:
I don't know how I missed that Renovate Bot had submitted a PR to update `@types/graphql` (#579). I'll close this PR in favor of that one. I would still be in favor of making `@types/graphql` a `devDependency` in all packages, as discussed above.

That said, #579 has been open since March 28. I don't think I'm alone by hoping it gets merged ASAP.

victorandree:
New PR which builds on #579 and fixes the build failure: #743.

johannpinson:
Hello @victorandree 
Any change to see a release soon with this update about `@types/graphql`?
I have some types conflicts with Apollo Server 2 and other packages which needs the lastest version but Apollo-link is always depending of the v0.12.6.

Thanks,

victorandree:
@johannpinson I'm not affiliated with Apollo and, like you, I'm waiting for a project maintainer to merge and release this. I don't know how to get their attention, however. @hwillson seems to have approved previous pull requests -- any chance you see this?

victorandree:
PRs merged to master... now we just wait for a new release!

PR #638, opened 2018-05-31
Bump @types/graphql dependency to 0.13.1

radar:

I'm seeing an issue with my build where it cannot pull down `@types/graphql@0.12.6`, and I think that is due to an upstream NPM issue. But it _can_ pull down `@types/graphql@0.13.1`. But because I use apollo-link, it is "locked" to using 0.12.6.

Bumping this version doesn't do any harm _and_ would fix my build issue.

### DIFF
--- a/packages/apollo-link/package.json
+++ b/packages/apollo-link/package.json
@@ -38,7 +38,7 @@
     "watch": "tsc -w -p . & rollup -c -w"
   },
   "dependencies": {
-    "@types/graphql": "0.12.6",
+    "@types/graphql": "0.13.1",
     "apollo-utilities": "^1.0.0",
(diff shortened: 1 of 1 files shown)


lukespice:
Duplicates https://github.com/apollographql/apollo-link/pull/634

andrewdushane:
I have CircleCI builds failing due to this same issue, thanks for the PR.

radar:
Looks like this is now fixed anyway: https://github.com/apollographql/apollo-link/pull/638#issuecomment-393385690

radar:
Looks like this is now fixed anyway: https://github.com/npm/registry/issues/340#event-1655013040

PR #641, opened 2018-06-04
Move @types/graphql to devDependencies in apollo-link

### LINKED ISSUE #601: Two different types with this name exist errors due to mismatch @types/graphql

amacleay:
Fixes #601 

Consumers of this repo should not implicitly get @types/graphql. This
can cause issues for consumers using different versions of the repo,
since they may have a different version of the type definitions which is
incompatible.

TODO:

- [n/a] Make sure all of new logic is covered by tests and passes linting

**Pull Request Labels**

- [ ] feature
- [ ] blocking
- [ ] docs


amacleay:
Identical to and fixed by #735 (thanks @benwaffle !)

PR #642, opened 2018-06-05
Support retrying failed request in error handler

clarencenpy:
This PR adds support for retrying failed requests from within ErrorLink's error handler. One use case that this could support is re-authenticating expired tokens on authentication failure, and retrying the request with new tokens. 

TODO:

**Pull Request Labels**

- [x] feature
- [ ] blocking
- [ ] docs

### DIFF
--- a/packages/apollo-link-error/CHANGELOG.md
+++ b/packages/apollo-link-error/CHANGELOG.md
@@ -1,8 +1,6 @@
 # Change log
 
 ### vNEXT
-
-### 1.0.10
 - Pass `forward` into error handler for ErrorLink to support retrying a failed request 
 
 ### 1.0.9
--- a/packages/apollo-link-error/CHANGELOG.md
+++ b/packages/apollo-link-error/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### vNEXT
 
+### 1.0.10
+- Pass `forward` into error handler for ErrorLink to support retrying a failed request 
+
 ### 1.0.9
 - Correct return type to FetchResult [#600](https://github.com/apollographql/apollo-link/pull/600)
 
--- a/packages/apollo-link-error/README.md
+++ b/packages/apollo-link-error/README.md
@@ -74,6 +74,11 @@ onError(({ graphQLErrors, networkError, operation, forward }) => {
 );
 ```
 
+Here is a diagram of how the request flow looks like now: 
+![Diagram of request flow after retrying in error links](https://imgur.com/a/pWy8G4k)
+
+One caveat is that the errors from the new response from retrying the request does not get passed into the error handler again. This helps to avoid being trapped in an endless request loop when you call forward() in your error handler.
+
 
 <h2 id="ignoring-errors">Ignoring errors</h2>
 
--- a/packages/apollo-link-error/README.md
+++ b/packages/apollo-link-error/README.md
@@ -66,7 +66,9 @@ onError(({ graphQLErrors, networkError, operation, forward }) => {
     }
     if (networkError) {
       console.log(`[Network error]: ${networkError}`);
-      // no need to retry if there was a network error
+      // if you would also like to retry automatically on 
+      // network errors, we recommend that you use 
+      // apollo-link-retry
     }
   }
 );
--- a/packages/apollo-link-error/README.md
+++ b/packages/apollo-link-error/README.md
@@ -29,12 +29,49 @@ Error Link takes a function that is called in the event of an error. This functi
 * `response`: The result returned from lower down in the link chain
 * `graphQLErrors`: An array of errors from the GraphQL endpoint
 * `networkError`: Any error during the link execution or server response, that wasn't delivered as part of the `errors` field in the GraphQL result
+* `forward`: A reference to the next link in the chain. Calling `return forward(operation)` in the callback will retry the request, returning a new observable for the upstream link to subscribe to. 
+
+Returns: `Observable<FetchResult> | void` The error callback can optionally return an observable from calling `forward(operation)` if it wants to retry the request. It should not return anything else. 
 
 <h2 id="error-categories">Error categorization</h2>
 
 An error is passed as a `networkError` if a link further down the chain called the `error` callback on the observable. In most cases, `graphQLErrors` is the `errors` field of the result from the last `next` call.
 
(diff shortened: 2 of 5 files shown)


packages/apollo-link-error/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### vNEXT
 
+### 1.0.10
+- Pass `forward` into error handler for ErrorLink to support retrying a failed request
```
evans: Normally we place the changelog entry under `vNext` and then when we publish, we move them to the appropriate place. This is because the next version could be a minor version, rather than a patch, depending on the features that we add.

For a deeper look into how we normally version things, check out the [semver spec](https://semver.org/)

packages/apollo-link-error/src/__tests__/index.ts
```diff
@@ -232,6 +233,68 @@ describe('error handling', () => {
...
+      }
+    `;
+
+    let called;
+    let retryCalled;
+    const errorLink = onError(
+      ({ graphQLErrors, response, operation, forward }) => {
+        if (graphQLErrors.length > 0) {
+          called = true;
+          expect(graphQLErrors[0].message).toBe('resolver blew up');
```
evans: Let's surround these expects in a try -> catch -> done.fail, so that if the expect fail, then we get a good message

packages/apollo-link-error/src/index.ts
```diff
@@ -30,36 +31,67 @@ export const onError = (errorHandler: ErrorHandler): ApolloLink => {
...
               operation,
               networkError,
               //Network errors can return GraphQL errors on for example a 403
               graphQLErrors: networkError.result && networkError.result.errors,
+              forward,
             });
+            if (retryForward) {
+              retrying = true;
+              retrySub = retryForward.subscribe({
+                next: result => observer.next(result),
```
evans: We should be able to use the same pattern here and below as we did for `complete`. `observer.next.bind(observer)`.

packages/apollo-link-error/src/index.ts
```diff
@@ -30,36 +31,67 @@ export const onError = (errorHandler: ErrorHandler): ApolloLink => {
...
+              });
+              return;
+            }
             observer.error(networkError);
           },
-          complete: observer.complete.bind(observer),
+          complete: () => {
+            // disable the previous sub from calling complete on observable
+            // if retry is in flight.
+            if (!retrying) {
```
evans: nice catch! Can we reuse `retrySubscription` instead of defining another variable?

packages/apollo-link-error/src/index.ts
```diff
@@ -30,36 +31,67 @@ export const onError = (errorHandler: ErrorHandler): ApolloLink => {
...
       let sub;
+      let retrySub;
+      let retrying = false;
+
       try {
         sub = forward(operation).subscribe({
           next: result => {
             if (result.errors) {
-              errorHandler({
+              const retryForward = errorHandler({
```
evans: `retryForward` seems like a confusing name. Something more aligned with what the return result is would be appropriate. Maybe `overridingObservable` or `retriedResult`?

packages/apollo-link-error/README.md
```diff
@@ -29,13 +29,50 @@ Error Link takes a function that is called in the event of an error. This functi
...
+              },
+            });
+            // retry the request, returning the new observable
+            return forward(operation);
+        }
+      }
+    }
+    if (networkError) {
+      console.log(`[Network error]: ${networkError}`);
+      // no need to retry if there was a network error
```
evans: Adding a sentence to this comment about using apollo-link-retry to deal with `networkError`s would help users understand the distinction between link-error and link-retry

evans:
It looks like codecov is complaining about the coverage. Check out this https://codecov.io/gh/apollographql/apollo-link/pull/642/src/packages/apollo-link-error/src/index.ts?before=packages/apollo-link-error/src/index.ts to see what changed

packages/apollo-link-error/CHANGELOG.md
clarencenpy: Fixed

packages/apollo-link-error/src/__tests__/index.ts
clarencenpy: Fixed

packages/apollo-link-error/src/index.ts
clarencenpy: fixed

packages/apollo-link-error/src/index.ts
clarencenpy: fixed

packages/apollo-link-error/src/index.ts
clarencenpy: fixed

packages/apollo-link-error/README.md
clarencenpy: fixed

packages/apollo-link-error/README.md
```diff
@@ -74,6 +74,11 @@ onError(({ graphQLErrors, networkError, operation, forward }) => {
 );
 ```
 
+Here is a diagram of how the request flow looks like now: 
+![Diagram of request flow after retrying in error links](https://imgur.com/a/pWy8G4k)
```
evans: This is so lit!

packages/apollo-link-error/src/__tests__/index.ts
```diff
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 import { ApolloLink, execute, Observable } from 'apollo-link';
-
 import { onError, ErrorLink } from '../';
+import {} from 'jest';
```
evans: I'm not sure what this import provides. It seems like something that should be caught with the right tsconfig. If we can remove it, then let's do it, otherwise we can leave a comment about why it's necessary.

packages/apollo-link-error/README.md
```diff
@@ -29,13 +29,57 @@ Error Link takes a function that is called in the event of an error. This functi
...
+      // if you would also like to retry automatically on 
+      // network errors, we recommend that you use 
+      // apollo-link-retry
+    }
+  }
+);
+```
+
+Here is a diagram of how the request flow looks like now: 
+![Diagram of request flow after retrying in error links](https://imgur.com/a/pWy8G4k)
```
evans: This is so lit!

packages/apollo-link-error/src/__tests__/index.ts
clarencenpy: Oh yeah totally! Im not sure if it is something specific to webstorm, but I had to add it otherwise there are a whole bunch of tslint. I wanted to remove it before committing! removing..

PR #643, opened 2018-06-05
Remove params directive option from docs

petetnt:
- [x] docs

I don't think the `params` option has ever existed outside of the [initial spec](https://github.com/apollographql/apollo-link-rest/blob/5d94527102bbd6886ba46dbff422f816f0f66aed/designs/initial.md), instead the params are automatically mapped from the docs.


stubailo:
Thanks!

PR #653, opened 2018-06-19
Remove broken link checker from this repository.

abernix:
This partially reverts commit ca9a72ea296a90706c8bcaaf0d02a597ece5b07c from
apollographql/apollo-link#597, since the link checking is not working.

As discussed in other repositories, such as `apollo-client` and
`apollo-server`, it would be better to hoist this checking up to a higher,
site-wide location or introduce this functionality into our docs system
itself, rather than introducing it on individual repositories.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
           key: dependency-cache-{{ checksum "./docs/package.json" }}
           paths:
             - ./docs/node_modules
-      - run: cd docs && npm run ci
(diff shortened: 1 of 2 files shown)


abernix:
Relates to: https://github.com/apollographql/apollo-client/pull/3313#issuecomment-395143574 and https://github.com/apollographql/engine-docs/pull/138.

abernix:
This is the first status-check-green ✅ PR in this link repository in quite some time! 🎉

PR #714, opened 2018-07-13
Update HttpLink to accept multipart response from deferred requests

clarencenpy:


packages/apollo-link-http/package.json
```diff
@@ -40,14 +40,16 @@
   },
   "dependencies": {
     "apollo-link": "^1.2.2",
-    "apollo-link-http-common": "^0.2.4"
+    "apollo-link-http-common": "^0.2.4",
+    "text-encoding": "^0.6.4"
```
hwillson: Whoa - `text-encoding` is massive! 😱 [189k minified and gzipped](https://bundlephobia.com/result?p=text-encoding@0.6.4) - ouch! Is there another package we can use instead, or some way we can reduce this drastically?

clarencenpy: Yeah I know ><
The good news is TextDecoder is already present as a global on [most browsers](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder#Browser_compatibility), but with enough of those that don't support it to require a polyfill. We may also need a polyfill for ReadableStreams API, but I will research better options!

PR #783, opened 2018-09-06
Fix the typing of data attribute in FetchResult

mvestergaard:
When invoking a mutation directly via `apollo-client`, it returns a [`FetchResult<T>`](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/ApolloClient.ts#L271)

However the first argument is not used for the `data` attribute, instead it types the `context` attribute.  
So with this change the first argument instead specifies the type of `data`, instead of `context`.

/label enhancement

### DIFF
--- a/packages/apollo-link/CHANGELOG.md
+++ b/packages/apollo-link/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change log
 
 ### vNEXT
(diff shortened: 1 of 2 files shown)


rosskevin:
This was superseded by #804

JoviDeCroock:
I think it's worth closing this since the work has already been done in #804

PR #814, opened 2018-10-01
attach raw response to context when using batch links instead of just body

### LINKED ISSUE #550: BatchHttpLink does not add HTTP Response to context
**Expected Behavior**
Consistent behavior with HttpLink by exposing the complete HTTP response through `context.response`

**Actual Behavior**
`context.response` is the response body and not the full HTTP response

**A _simple_ reproduction**
https://codesandbox.io/s/0q95olj22l
This will make a simple request to the SWAPI (https://graphiql.graphcms.com/simple/v1/swapi) using both HttpLink and BatchHttpLink and log the context.

**Issue Labels**

- [x] has-reproduction
- [ ] feature
- [ ] docs
- [ ] blocking
- [ ] good first issue

### LINKED ISSUE #631: BatchHttpLink does not provide exposed response headers
I have this afterware that was working great till we implemented `import { BatchHttpLink } from 'apollo-link-batch-http'`. Now I no longer have access to the response headers. 
```
//const httpLink = createHttpLink(
//  { uri: GRAPHQL_ENDPOINT, credentials: 'same-origin' }
//)

 const httpLink = new BatchHttpLink({
   uri: GRAPHQL_ENDPOINT,
   credentials: 'same-origin'
 })

const afterwareLink = new ApolloLink((operation, forward) => {
  return forward(operation).map(response => {
    const context = operation.getContext()
    const {
      response: { headers }
    } = context

    if (headers) {
      const logout = headers.get('logout')

      if (logout) {
        LocalStorage.removeItem('access_token')
      }
    }
    return response
  })
})

const link = ApolloLink.from([
  catchErrorLink,
  middlewareLink,
  afterwareLink,
  httpLink
])
```
The only property that lives on response is "data" holding query results. 

**Expected Behavior**

I would like the  BatchHttpLink to return response headers that are exposed by the server.  My server is properly exposing it.

03eltond:
This bugfix addresses the issue of the raw response not being set on context in `apollo-link-batch-http`, like it does in `apollo-link-http`.  This is especially helpful for inspecting headers coming back from the GraphQL server.

Since the request is not available by the time it gets to `apollo-link-batch`, the `setContext` call was moved into `apollo-link-batch-http` in a similar fashion to how `apollo-link-http` does it.

Fixes #550.
Fixes #631.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @hwillson @benjamn
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -23,18 +23,3 @@ Places to look for information include your browser console, server console, and
 example: A Codesandbox or GitHub repository that anyone can clone to observe the problem
 -->
 
-**Issue Labels**
-
-<!--
-While not necessary, you can help organize our issues by labeling this issue when you open it.  To add a label automatically, simply [x] mark the appropriate box below:
--->
-
-- [ ] has-reproduction
-- [ ] feature
-- [ ] docs
-- [ ] blocking
-- [ ] good first issue
-
-<!--
-To add a label not listed above, simply place `/label another-label-name` on a line by itself.
--->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,16 +10,3 @@ TODO:
 - [ ] Make sure all of new logic is covered by tests and passes linting
 - [ ] Update CHANGELOG.md with your change
 
-**Pull Request Labels**
-
-<!--
-While not necessary, you can help organize our pull requests by labeling this issue when you open it.  To add a label automatically, simply [x] mark the appropriate box below:
--->
-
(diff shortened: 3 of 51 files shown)


chrispage1:
Is there any way this pull request can be merged? I have just run into this issue and discovered numerous issues where this has been a problem. I've tested @03eltond changes and they do indeed provide the required fix.

As @roupen highlighted in Issue #231 after it had been closed, this issue still persists.

boom:
Bump again. Running into this same issue, switched to batch http link and cannot parse a header we were parsing before when using http link.

roupen:
@boom for what it's worth if you need a workaround until it gets merged I had luck checking the response headers in a custom fetch:

```
const customFetch = (uri, options) => {
  let promise = fetch(uri, options)
  promise
    .then(response => {
      let refreshToken = response.headers.get('Refresh-Token')
      if (refreshToken) {
        ctx.store.dispatch('doSomething', refreshToken)
      }
      return response
    })
    .catch(e => {
      throw e
    })
  return promise
}

const httpLink = new BatchHttpLink({ uri: process.env.GRAPHQL_URL, fetch: customFetch })
```

PR #823, opened 2018-10-09
Fix missing genenic type in types definition

NateScarlet:
```Typescript
apolloClient.mutate<T>().then(result => result)
```
has no type hint on `result.data` before fix.

TODO:

**Pull Request Labels**

- [ ] feature
- [ ] blocking
- [ ] docs


NateScarlet:
I found out `C` is for context
maybe i should use something like 
```TypeScript
apolloClient.mutate().then((result:T) => result)
```

PR #831, opened 2018-10-16
Use `ssrMode` instead of `ssr` as an option for ApolloClient

ohbarye:
This PR just fixes a document to follow the newer Apolloclient.

ref https://www.apollographql.com/docs/link/links/schema.html

image

TODO:

I think all of them are not necessary since this is just a fix for the document.

**Pull Request Labels**

- [ ] feature
- [ ] blocking
- [x] docs


benjamn:
Thanks @ohbarye!

PR #832, opened 2018-10-18
Fix SubscriptionClient detection

### LINKED ISSUE #773: [apollo-link-ws] Uncaught DOMException: Failed to construct 'WebSocket': The URL 'undefined' is invalid.
Hi,

The issue is described in https://github.com/prisma/graphql-playground/issues/600 but I think it should be opened here.

To summarise:

```js
// Using the SubscriptionClient syntax stopped working for some reason
const subscriptionClient = new SubscriptionClient(GRAPHQL_ENDPOINT, {
  reconnect: true,
  connectionParams: {
    authToken: access_token,
  },
});
const link = new WebSocketLink(subscriptionClient);  // throws


// This one still works
const link = new WebSocketLink({
  uri: GRAPHQL_ENDPOINT,
  options: {
    reconnect: true,
    connectionParams: {
      authToken: access_token,
    },
  }
});
```
Please read the original issue above; if you can't reproduce the problem I'll try to setup a CodeSandbox ASAP. 

Thanks in advance! ❤️

nuragic:
Close #773.

`/label bug`

### DIFF
--- a/packages/apollo-link-ws/src/webSocketLink.ts
+++ b/packages/apollo-link-ws/src/webSocketLink.ts
@@ -35,11 +35,14 @@ export class WebSocketLink extends ApolloLink {
   ) {
     super();
 
-    if ((<WebSocketLink.Configuration>paramsOrClient).uri !== undefined) {
+    if ((<WebSocketLink.Configuration>paramsOrClient).uri) {
+      const { uri, options, webSocketImpl } = <WebSocketLink.Configuration>(
+        paramsOrClient
+      );
       this.subscriptionClient = new SubscriptionClient(
-        (<WebSocketLink.Configuration>paramsOrClient).uri,
-        (<WebSocketLink.Configuration>paramsOrClient).options,
-        (<WebSocketLink.Configuration>paramsOrClient).webSocketImpl,
+        uri,
+        options,
+        webSocketImpl,
       );
     } else {
       this.subscriptionClient = <SubscriptionClient>paramsOrClient;
--- a/packages/apollo-link-ws/src/__tests__/webSocketLink.ts
+++ b/packages/apollo-link-ws/src/__tests__/webSocketLink.ts
@@ -1,4 +1,5 @@
 import { SubscriptionClient } from 'subscriptions-transport-ws';
+import { SubscriptionClient as SubscriptionClient2 } from 'subscriptions-transport-ws';
 import { Observable, execute } from 'apollo-link';
 import { ExecutionResult } from 'graphql';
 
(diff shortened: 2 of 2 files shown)


packages/apollo-link-ws/src/webSocketLink.ts
```diff
@@ -35,14 +35,14 @@ export class WebSocketLink extends ApolloLink {
   ) {
     super();
 
-    if (paramsOrClient instanceof SubscriptionClient) {
-      this.subscriptionClient = paramsOrClient;
-    } else {
+    if ((<WebSocketLink.Configuration>paramsOrClient).uri !== undefined) {
```
JoviDeCroock: Isn't it possible to do this in one step? Like cast it once and not every time you use it? Just asking

nuragic: Hi Jovi! Glad you're reviewing this. First of all, just to be clear: I'm not a TypeScript expert, neither a TS user... This is one of my first times writing TS code. 😄
Regarding your question: I'm not sure if we can do that... neither if it would change the real JS output, which is, right now:

```js
var WebSocketLink = (function (_super) {
    __extends(WebSocketLink, _super);
    function WebSocketLink(paramsOrClient) {
        var _this = _super.call(this) || this;
        if (paramsOrClient.uri !== undefined) {
            _this.subscriptionClient = new subscriptions_transport_ws_1.SubscriptionClient(paramsOrClient.uri, paramsOrClient.options, paramsOrClient.webSocketImpl);
        }
        else {
            _this.subscriptionClient = paramsOrClient;
        }
        return _this;
    }
    WebSocketLink.prototype.request = function (operation) {
        return this.subscriptionClient.request(operation);
    };
    return WebSocketLink;
}(apollo_link_1.ApolloLink));
```

However, if you can confirm that it could be changed, I'd change it indeed. Also feel free to change anything!

JoviDeCroock: I'll try to free up some time this afternoon to look at it, I know the JS out put does not change it just reads pretty hard casting it multiple times.

JoviDeCroock:
What could help is a test where the previous scenario would fail, if you could hypothetically provide that case or implement it, it would be of great help for me to understand the problem.

nuragic:
The failing scenario is described here: https://github.com/apollographql/apollo-link/issues/773#issuecomment-420171145

As multiple versions of the lib could be installed (not sure if this is ok or it shall never happen; the thing is that it happens to many people), then the previous `instanceof` check could fail.

What I'm proposing with this PR is basically to replace the `instanceof` check because it's not reliable as it doesn't work cross realm.

packages/apollo-link-ws/src/webSocketLink.ts
nuragic: In vanilla JS I'd probably write something like this:

```ts
if ((<WebSocketLink.Configuration>paramsOrClient).uri) {
  const { uri, options, webSocketImpl } = (<WebSocketLink.Configuration>paramsOrClient);
  this.subscriptionClient = new SubscriptionClient(uri, options, webSocketImpl);
} else {
  this.subscriptionClient = <SubscriptionClient>paramsOrClient;
}
```

I've tried it and it works / compile as expected. Could work for you?

JoviDeCroock: Yes, I'll take it up to the apollo-maintainers tomorrow. Thanks for the quick feedback

nuragic:
Just pushed a new commit, I hope those tests would be enough 😄

yoshiakis:
@JoviDeCroock Would you review this again if you had time?

PR #850, opened 2018-11-01
Fix installation directions

petetnt:
This PR fixes the `link-rest` docs which don't list the required peerDepencendies, causing issues like https://github.com/nihakue/apollo-rest-link-minimal-repro

The installation directions are correct in https://github.com/apollographql/apollo-link-rest#installation

- [ ] feature
- [ ] blocking
- [x] docs


evans:
@petetnt Thank you!

PR #880, opened 2018-11-29
handle direct fetch abort

rezof:
- fixes [issue](https://github.com/apollographql/apollo-client/issues/4150)
- test already covered, but i can write a specific test for it if needed

TODO:

- [X] Make sure all of new logic is covered by tests and passes linting

### DIFF
--- a/packages/apollo-link-http/src/httpLink.ts
+++ b/packages/apollo-link-http/src/httpLink.ts
@@ -144,8 +144,8 @@ export const createHttpLink = (linkOptions: HttpLink.Options = {}) => {
           return result;
         })
         .catch(err => {
-          // fetch was cancelled so its already been cleaned up in the unsubscribe
-          if (err.name === 'AbortError') return;
+          // exit if fetch was cancelled from the observer unsubscribing
+          if (err.name === 'AbortError' && observer.closed) return;
           // if it is a network error, BUT there is graphql result info
           // fire the next observer before calling error
           // this gives apollo-client (and react-apollo) the `graphqlErrors` and `networErrors`


JMadelaine:
I've run into this issue. When will this change be implemented as it is holding my project back?

JoviDeCroock:
> I've run into this issue. When will this change be implemented as it is holding my project back?

When this test is implemented it will probably be pushed forward.

rezof:
@hwillson @JoviDeCroock test added

Optarion:
Any news on this?

packages/apollo-link-http/src/httpLink.ts
```diff
@@ -144,8 +144,8 @@ export const createHttpLink = (linkOptions: HttpLink.Options = {}) => {
           return result;
         })
         .catch(err => {
-          // fetch was cancelled so its already been cleaned up in the unsubscribe
-          if (err.name === 'AbortError') return;
+          // exit if fetch was cancelled from the observer unsubscribing
+          if (err.name === 'AbortError' && observer.closed) return;
```
helfer: Think you don't even need to check what type of error it is. If nobody is listening (`observer.closed`), you can return here.

helfer:
FWIW, I think this PR makes sense, but it isn't a solution to the original issue. Using an AbortController to reach into the link stack and abort a request instead of unsubscribing breaks the abstraction. If you continue trying to work around it by putting in ifs and elses here and there, the code will become bloated and much more complicated, and you'll likely introduce more bugs and corner-cases along the way.

I think the most logical way to go about this would be to make the promise returned by `client.query` cancelable (if it isn't already), and then propagating that cancellation up the stack. Presumably this is already what happens with `watchQuery` anyway when you unsubscribe?

packages/apollo-link-http/src/httpLink.ts
helfer: In fact, you shouldn't care what type of error it is. As long as someone is subscribed, they need to find out about this error, otherwise you're just leaving them hanging.

PS: As you probably figured out, the query deduplicator never finding out that the fetch was aborted is what's causing subsequent requests to hang.

mo:
It would be nice of useLazyQuery() returned a "cancel" function.

PR #905, opened 2019-01-09
Return null when headers are not parsed properly

steffansluis:
To indicate an incomplete multipart.

### DIFF
--- a/packages/apollo-link-http/src/httpLink.ts
+++ b/packages/apollo-link-http/src/httpLink.ts
@@ -66,6 +66,7 @@ function parseMultipartHTTP(plaintext: string): FetchResult[] | null {
     if (part.length) {
       let partArr = part.split('\r\n\r\n');
       if (!partArr || partArr.length !== 2) {
+        return null;
       }
 
       // Read the Content-Length header, which must be included in the response


JoviDeCroock:
Can the problem this PR fixes be tested, if so it would be awesome if we could implement this test so we don't break it in the near future.

steffansluis:
@JoviDeCroock I agree that would be a good idea, however note that the fix pertains to the branch `defer-support`, which itself has not been merged (#714) and does not have any tests for the updates yet. I'm not sure what the status of that PR is in general, but I've been looking into using this feature and I happened to notice this was causing errors. If the still actively being constructed, I'd be happy to add some tests. However, if the code is going stale I don't really see the point. I'd be happy to add tests anyway however. Maybe @clarencenpy can shed some light on the status of the defer-support?

JoviDeCroock:
Hmm, is this `defer-support` a big need? Since this seems really stale?

steffansluis:
The feature is really nice, although the way it's implemented it somewhat experimental, it works quite well. It basically allows you to side-load any part of your data for free. This can greatly improve the response time on request that require a bunch of nested data, since it allows you to start rendering as soon as you have the bare minimum of the required data. Check out [the demo app](https://github.com/apollographql/defer-demo/tree/master/client) if you want to see it in action.

Issue #910, opened 2019-01-16
apollo-link 1.2.5 / 1.2.6 causes typescript compile errors


gaberudy:
Introduced with this PR:
https://github.com/apollographql/apollo-link/pull/804

The `ExecutionResult<TData>` throws errors in modern typescript compilers.

My typescript error:
```

../node_modules/apollo-link/lib/types.d.ts:22:56 - error TS2315: Type 'ExecutionResult' is not generic.

22 }, C = Record<string, any>, E = Record<string, any>> = ExecutionResult<TData> & {
                                                          ~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

rosskevin:
@syndesis found he was using @types/graphql 0.12.6 which is 10 month old.  https://www.npmjs.com/package/@types/graphql

Looking to see the min version requirement now.

gaberudy:
Fair enough if newer graphql lib is needed, I also may be on an older version due to other library dependencies.

rosskevin:
@gaberudy we are chatting in #contributing right now on slack if you want to join

rosskevin:
@gaberudy are you also using apollo-link-dedup?

eknowles:
I have this same issue when importing `graphql-tools` package 😢

syndesis:
You can work around this issue with a `resolutions` section like this in your package.json:

```
"resolutions": {
    "apollo-link-dedup": "^1.0.13",
    "@types/graphql": "^14.0.0"
  }
```

JoviDeCroock:
Closing this since it is resolved, if anyone still has an issue after the proposed fix with resolutions feel free to reopen.

darmie:
> You can work around this issue with a `resolutions` section like this in your package.json:
> 
> ```
> "resolutions": {
>     "apollo-link-dedup": "^1.0.13",
>     "@types/graphql": "^14.0.0"
>   }
> ```

Unfortunately, this did not fix the problem for me

freehuoshan:
this issue still didn't resolved

bbugh:
This is happening for me after upgrading to the latest packages. Can we get this reopened please?

ShaharDadon:
I have the same issue. 

I've already tried: 
- resolution solution.
- upgrading apollo & graphql

any other suggestions?

qin-guan:
any fixes for this issue?

PowerKiKi:
[...]

PR #804, opened 2018-09-24
Expose generics from graphql ExecutionResult on FetchResult

rosskevin:
TODO:

- [ ] Make sure all of new logic is covered by tests and passes linting
- [x] Update CHANGELOG.md with your change

**Pull Request Labels**

- [ ] feature
- [x] blocking
- [ ] docs

Related to Jul 3 change in `@types/graphql`, update the `FetchResult` generics to strongly type the data.  I added `blocking` because [it prevents strong typing in usage of `apollo-client`](https://github.com/apollographql/apollo-client/issues/2795#issuecomment-424124181).

See: 
- https://github.com/DefinitelyTyped/DefinitelyTyped/commit/deb7d8beaff5d43da749c5bc0f2020cdcaaabb6b#diff-1a92e4f388a3da0eb35da5e193f02172
- https://github.com/apollographql/apollo-client/issues/2795#issuecomment-424124181

PR #911, opened 2019-01-17
Fix for issue apollo-link #910

syndesis:
Removed type from ExecutionResult to pass linting

### DIFF
--- a/packages/apollo-link/src/types.ts
+++ b/packages/apollo-link/src/types.ts
@@ -24,7 +24,7 @@ export type FetchResult<
   TData = { [key: string]: any },
   C = Record<string, any>,
   E = Record<string, any>
-> = ExecutionResult<TData> & {
+> = ExecutionResult & {
   extensions?: E;
   context?: C;
 };


rosskevin:
We should not untype a result, if the types has changed we should look at correcting it.

rosskevin:
Wow! the result is now untyped so this is the correct fix.  The updated code is https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L114-L117:

```tsx
/**
 * The result of GraphQL execution.
 *
 *   - `errors` is included when any errors occurred as a non-empty array.
 *   - `data` is the result of a successful execution of the query.
 */
export type ExecutionResult = {
  errors?: $ReadOnlyArray<GraphQLError>,
  data?: ObjMap<mixed>,
};
```

which is a step backwards and will introduce a problem code in apollo-client.  It looks like apollo-client will have to cast to `any` and then cast to a result.  

/cc @danilobuerger you might look at casting/correcting this inside apollo-client. I can't see a reason or find the commit where the generic was removed in https://github.com/graphql/graphql-js/commits/master/src/execution/execute.js but it is clearly gone now.  Since this is two steps removed from apollo-client, perhaps it is best to just cast it there?

@hwillson this is good to merge and will need a patch release to avoid breakage with `graphql` dependency.

rosskevin:
Looks like my original PR #804 was based on `@types/graphql` - still looking - need to check definitelytyped....

rosskevin:
Ok, looks like the code is still on DefinitelyTyped, so this should not be needed.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/graphql/execution/execute.d.ts#L43

```tsx
/**
 * The result of GraphQL execution.
 *
 *   - `errors` is included when any errors occurred as a non-empty array.
 *   - `data` is the result of a successful execution of the query.
 */
export interface ExecutionResult<TData = ExecutionResultDataDefault> {
    errors?: ReadonlyArray<GraphQLError>;
    data?: TData;
}
```

gaberudy:
See the discussion of adding types to the data result here for DefinitelyTyped:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26763

I think not being type compatible with graphql-js is a strong negative though and caused my issue.

syndesis:
Cancelling per updates from @rosskevin

PR #916, opened 2019-01-19
Support subscriptions in apollo-link-schema.

fubhy:
SchemaLink currently doesn't support subscriptions. This pull request adds that.

I understand that this is probably not a very common so I'd like to explain my use-case:

We are running a hybrid desktop & web app with environment specific differences in session handling and security. The application is a Ethereum DApp, hence this needed complexity (local self-hosted blockchain nodes vs. shared centrally hosted nodes for easy trial access on the web). We abstracted all of this logic into the data layer (graphql & apollo) by using different adapters/transports based on the environment that the application runs in. E.g. in the web version, we use a stitched schema with a partly centralized schema and a partly decentralized schema. The first part runs in a webworker locally within the browser and handles the session specific parts of the schema, the second part runs on a server. In case of the electron app, the whole graphql schema runs locally in the main process with a custom transport through the electron IPC channel.

In both cases, we use a fork (with subscription support as added by this pull request) of the schema link as the final "adapter" for executing the queries (either after passing them to the webworker or through IPC).


jedwards1211:
I'd love to see this merged soon, this would make some of my integration testing a lot simpler.

jedwards1211:
In my case I want to do some "integration" testing on my React views with `enzyme` and `apollo-link-schema`, but I use GraphQL subscriptions so atm I would have to start a server and connect to it via a client.

Issue #259, opened 2017-11-17
apollo-link-dedup - subscribeToMore will unsubscribe after first data | bug


seeden:
I found that I am automatically unsubscribed when I receive first data with subscribeToMore.
The problem is in the apollo-link-dedup (https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-dedup/src/dedupLink.ts#L73)

More info about this issue you can find here:
https://github.com/apollographql/apollo-client/issues/2581

```js
    "apollo-cache-inmemory": "^1.1.1",
    "apollo-client": "^2.0.3",
    "apollo-link": "^1.0.3",
    "apollo-link-http": "^1.2.0",
    "apollo-link-batch-http": "^1.0.1",
    "apollo-link-retry": "^1.0.1",
    "apollo-link-dedup": "^1.0.2",
    "apollo-link-ws": "^1.0.1",
    "apollo-utilities": "^1.0.2",
```

azazdeaz:
As i see it closes the regular queries as well after the first response.

If i roll back to `apollo-link-dedup@1.0.0` everything works properly again.

I do not fully understand the logic of dedupLink but the root of the issue is maybe [here](https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-dedup/src/dedupLink.ts#L61) where it calls observer.complete after the first next() call.

cc @jbaxleyiii

seeden:
I tried apollo-link-dedup@1.0.0 and it is working for me too.

jbaxleyiii:
@seeden that is no good at all! I'll see about writing a failing test, would you be able to help with that?

seeden:
@jbaxleyiii  failing test case is very simple. Use graphql decorator from react-apollo and call subscribeToMore in the componentDidMount when data.loaded is true. You will be automatically unsubscribed when you receive first data from subscription.

azazdeaz:
@jbaxleyiii I'm trying to understand how it should work to write a failing test. Can you help me why does it have to `complete` all the subscribers ([here](https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-dedup/src/dedupLink.ts#L61)) when the first piece of data arrives ?

schavaLogi:
Any update on this, this is effecting apolli-link-ws subscriptions

seeden:
any update ?

typical000:
After upgrading all apollo* dependencies to latest version solved the issue (in my case :))

PR #956, opened 2019-02-22
Fix: ensure that all observers canceled when closing fetch

HugeSkull:
Fixes: #259

### DIFF
--- a/packages/apollo-link-dedup/src/__tests__/dedupLink.ts
+++ b/packages/apollo-link-dedup/src/__tests__/dedupLink.ts
@@ -268,3 +268,45 @@ describe('DedupLink', () => {
     expect(unsubscribed).toBe(true);
   });
 });
+
+it(`should be wait all observers unsubscribe`, () => {
+  const document: DocumentNode = gql`
+    query test1($x: String) {
+      test(x: $x)
+    }
+  `;
+  const variables1 = { x: 'Hello World' };
+  const variables2 = { x: 'Hello World' };
+
+  const request1: GraphQLRequest = {
+    query: document,
+    variables: variables1,
+    operationName: getOperationName(document),
+  };
+
+  const request2: GraphQLRequest = {
+    query: document,
+    variables: variables2,
+    operationName: getOperationName(document),
+  };
+
+  let unsubscribed = false;
+  const deduper = ApolloLink.from([
+    new DedupLink(),
+    new ApolloLink(() => {
+      return new Observable(() => {
+        return () => {
+          unsubscribed = true;
+        };
+      });
+    }),
+  ]);
+
+  const sub1 = execute(deduper, request1).subscribe({});
+  const sub2 = execute(deduper, request2).subscribe({});
+
+  sub1.unsubscribe();
+  expect(unsubscribed).toBe(false);
+  sub2.unsubscribe();
+  expect(unsubscribed).toBe(true);
+});
--- a/packages/apollo-link-dedup/src/dedupLink.ts
+++ b/packages/apollo-link-dedup/src/dedupLink.ts
@@ -6,6 +6,8 @@ import {
   Observable,
 } from 'apollo-link';
 
+type Observer<T> = ZenObservable.Observer<T>;
+
 /*
  * Expects context to contain the forceFetch field if no dedup
  */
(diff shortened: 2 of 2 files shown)


packages/apollo-link-dedup/src/dedupLink.ts
```diff
@@ -14,7 +16,7 @@ export class DedupLink extends ApolloLink {
     string,
     Observable<FetchResult>
   > = new Map();
-  private subscribers: Map<string, any> = new Map();
+  private subscribers: Map<string, Array<Observer<FetchResult>>> = new Map();
```
benjamn: Why not `Map<string, Set<Observer<FetchResult>>>`?

HugeSkull: Hi @benjamn ! Glad you're reviewing this. 
That's a good suggestion. I'll revise it.

HugeSkull:
This PR and #984 are repeated.
Fine, I'll close this.I'd like to consult what's better about #984  than me. My PR was opened for 20 days, but it was merged after 4 hours. It would be great if you could give me some advice. @benjamn @hwillson

hwillson:
Sorry @HugeSkull! Don't close this - I'll review it shortly, and see if we can merge the approaches. Yesterday we needed to resolve the issue your PR and #984 both fix as soon as possible, and I took the first PR I noticed that addressed the issue. I should have taken a closer look at the outstanding PR list and considered your PR first. Thanks very much for your contribution - and sorry again!

HugeSkull:
HI @hwillson . I'm glad you can reply to me. You don't have to apologize for that. I have a responsibility too. I should be more proactive push on this PR.
This PR may not be able to merge, because my solution is basically the same as #984 . Maybe I can contribute some Typescript type definitions.

JoviDeCroock:
@HugeSkull, the solution is alike with the difference that your solution has no "destructive" elements on `next` and `error` that was the problem that occured this morning. There were situations where the observer already was deleted before the cleanup stage. This resulted in applications crashing without reason.

HugeSkull:
@JoviDeCroock , Sorry, I don't quite understand `destructive elements`. Are you referring to the fact that I don't have a null check [here?](https://github.com/apollographql/apollo-link/pull/956/files#diff-bbe7a318e5777466d1a4990ce7b90449R76) Under what circumstances will the observer be deleted in advance?

JoviDeCroock:
@HugeSkull I am referring to what is in master right now https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-dedup/src/dedupLink.ts#L47

This makes for scenario's where the disposer function crashes itself by trying to call `.delete` on a non-existent observer. I don't think this issue can occur with how you go about this issue.

HugeSkull:
@JoviDeCroock . Sorry, maybe I still don't quite understand what you mean. `this.subscribers` is a Map.  Deleting a non-existent key will cause the application to crash?  (And I moved the cleanup operation to the `cleanup` function.)

packages/apollo-link-dedup/src/__tests__/dedupLink.ts
```diff
@@ -268,3 +268,45 @@ describe('DedupLink', () => {
     expect(unsubscribed).toBe(true);
   });
 });
+
+it(`should be wait all observers unsubscribe`, () => {
```
benjamn: I added a very similar test in [this commit](https://github.com/apollographql/apollo-client/pull/4576/commits/3c40d984a78a1bc0fc6c7957d43b1132d9f8f85a#diff-0a0243849ea07e7f9b27a68afb1cd653), in case you want to check it out.

packages/apollo-link-dedup/src/dedupLink.ts
```diff
@@ -42,36 +45,38 @@ export class DedupLink extends ApolloLink {
...
             },
           });
         }
 
         return () => {
-          if (subscription) subscription.unsubscribe();
-          this.inFlightRequestObservables.delete(key);
+          const observers = this.subscribers.get(key);
+          observers.delete(observer);
+          if (observers.size === 0) {
```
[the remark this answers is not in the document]
HugeSkull: I think the length of the `observer` set and the number of times the `unsubscribe callback` is called should always be the same.So there will be no case where deleting the `non-existent key` causes the set to be empty.And the operation in the **else branch** depends on `observers.delete(observer)` Side effects, if put it in the **if condition**,it will be implicit and not semantic

packages/apollo-link-dedup/src/__tests__/dedupLink.ts
HugeSkull: This is great. :stuck_out_tongue_winking_eye:

PR #957, opened 2019-02-22
Add useContext option for dedup link

### LINKED ISSUE #920: Two request can get deduplicated if same query and vars but different headers
Two queries can get wrongly deduplicated if they are requesting the same query and vars to the same host, but with different HTTP headers. Perhaps we can include all or some request Headers into the `getKey` of the operation type, so that case is well covered and never the second request got hidden with the result of the first one.

https://github.com/apollographql/apollo-link/blob/089f71be5ea6a2ccd465751a500b16b97dbc23aa/packages/apollo-link/src/linkUtils.ts#L127

I do not know even how to implement this, but first: is this idea logic, reliable, and feasible??

Thanks,

Druotic:
Uses context config for uniqueness when deduping requests. Fixes #920

note: This is particularly useful for otherwise duplicate queries (same query/variables) that have different headers. For example, when you offload authorization details to a downstream service - you need separate/non-deduped requests with their own headers in each request.

TODO:

- [X] Make sure all of new logic is covered by tests and passes linting
- [] Update CHANGELOG.md with your change

### DIFF
--- a/packages/apollo-link-dedup/src/dedupLink.ts
+++ b/packages/apollo-link-dedup/src/dedupLink.ts
@@ -16,16 +16,26 @@ export class DedupLink extends ApolloLink {
   > = new Map();
   private subscribers: Map<string, any> = new Map();
 
+  constructor({ useContext = false }: { useContext?: boolean } = {}) {
+    super();
+    this.useContextForDedup = useContext;
+  }
+
   public request(
     operation: Operation,
     forward: NextLink,
   ): Observable<FetchResult> {
+    const context = operation.getContext();
     // sometimes we might not want to deduplicate a request, for example when we want to force fetch it.
-    if (operation.getContext().forceFetch) {
+    if (context.forceFetch) {
       return forward(operation);
     }
 
-    const key = operation.toKey();
+    let key = operation.toKey();
+    if (this.useContextForDedup) {
+      //may throw error if config not serializable
+      key = `${key}-${JSON.stringify(context)}`;
+    }
 
     const cleanup = operationKey => {
       this.inFlightRequestObservables.delete(operationKey);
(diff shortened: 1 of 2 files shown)


packages/apollo-link-dedup/src/dedupLink.ts
```diff
@@ -16,16 +16,26 @@ export class DedupLink extends ApolloLink {
...
-    if (operation.getContext().forceFetch) {
+    if (context.forceFetch) {
       return forward(operation);
     }
 
-    const key = operation.toKey();
+    let key = operation.toKey();
+    if (this.useContextForDedup) {
+      //may throw error if config not serializable
+      key = `${key}-${JSON.stringify(context)}`;
```
Druotic: Similar pattern/idea from here: https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-batch-http/src/batchHttpLink.ts#L207

Druotic:
Oops - Typescript! Update incoming.

Druotic:
@robermorales @benjamn @hwillson any thoughts on this change? I'm happy to address any feedback you might have.

Druotic:
We've been using this change in production for a couple of weeks now and it's worked great for us.

Druotic:
I pushed an update to resolve conflicts with master. Looking forward to this being released!

packages/apollo-link-dedup/src/dedupLink.ts
benjamn: Shouldn't this logic be moved into `operation.toKey`, so that we don't have multiple layers of key building?

Druotic: @benjamn I considered moving it there but I wasn't sure if that was the preferred approach since there was already some precedent for building the key in the specific package - e.g. the link above for `apollo-link-batch-http`. This approach also minimized the blast radius by being opt-in only.

Even though it feels like this should be the default behavior, if this is moved to `operation.toKey` with no opt-in mechanism, it technically becomes a breaking change. I suppose we could still check the context for some option, but that might seem a little strange as a consumer to specify on every request.

Would you rather it be slightly clunky for the sake of being opt-in or just have it be the default behavior?

JoviDeCroock:
Personally I don't see an issue moving filesize limit for dedup to 575/600B, do you @benjamn?

PR #959, opened 2019-02-26
Import tslib helpers from shared package rather than inlining.

benjamn:
Most Apollo applications use at least three link packages: `apollo-link`, `apollo-link-http-common`, and `apollo-link-http`.

These packages import helpers like `__extends`, `__assign`, and `__rest` from the `tslib` package.

If we inline these helpers in each package, we can control how much of the `tslib` package gets bundled (only what we use), but the helpers will be duplicated unnecessarily.

By importing the helpers from `tslib` as an external package, we can share one copy of the `tslib` package among all Apollo packages, and bundlers that understand ECMAScript `import` syntax can still prune the contents of the `tslib` package to include only the helpers that are used (throughout the whole application).


benjamn:
Thanks for the review @JoviDeCroock!

PR #965, opened 2019-03-04
Attempt to fix out-of-memory Circle CI failures.

benjamn:
We're still seeing intermittent out-of-memory failures in our Circle CI test (e.g. https://circleci.com/gh/apollographql/apollo-link/28075), which seem to have gotten worse since I merged https://github.com/apollographql/apollo-link/pull/959.

This PR is an attempt to simplify our npm scripts so that they don't use as much memory, by using `file:...` dependencies instead of `lerna link`, and simplifying the `prepare` scripts of the child packages. It may not be enough by itself, but I think it's a step in the right direction.

### DIFF
--- a/packages/apollo-link-batch-http/package.json
+++ b/packages/apollo-link-batch-http/package.json
@@ -28,7 +28,7 @@
     "filesize": "npm run build",
     "lint": "tslint -c \"../../tslint.json\" -p tsconfig.json -c ../../tslint.json src/*.ts",
     "prebuild": "npm run clean",
-    "prepublishOnly": "npm run clean && npm run build",
+    "prepare": "npm run build",
     "test": "npm run lint && jest",
     "watch": "tsc -w -p . & rollup -c -w"
   },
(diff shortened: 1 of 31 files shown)


packages/apollo-link-batch-http/package.json
```diff
@@ -52,7 +51,7 @@
     "object-to-querystring": "1.0.8",
     "proxyquire": "1.8.0",
     "rimraf": "2.6.3",
-    "rollup": "0.68.2",
+    "rollup": "^1.4.1",
```
JoviDeCroock: Won't the upgrade-bot complain about ^?

JoviDeCroock:
Sadly another out of memory :( maybe the task is too big idk?

packages/apollo-link-batch-http/package.json
benjamn: Ahh good point. Pinned: 4447dc99e0820614ba8eee11f1ea3b5ef7a78960

benjamn:
Lesson learned: third-party Rollup plugins can be pretty resource-intensive!

PR #966, opened 2019-03-04
Reduce bundlesize limits to reflect recent measurement changes.

benjamn:
Although it's possible that shipping ESM code via the "module" field of `package.json` may have reduced the size of these packages, the primary reason for the very large reduction in package sizes is that we're no longer bundling in the dependencies, so the new sizes reflect just the packages themselves.

Including dependencies in per-package size measurements is misleading because most apps use multiple `apollo-link-*` packages, so shared dependencies will be included only once in the bundle, and thus should not be counted as part of each individual package's size.

Fundamentally, measuring bundle size on a per-package basis doesn't make a whole lot of sense. Instead, we should use an example application that uses a typical/representative set of Apollo packages, and bundles everything together using a modern bundling tool.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,47 +25,47 @@
     {
       "name": "apollo-link",
       "path": "./packages/apollo-link/lib/bundle.min.js",
-      "maxSize": "3 Kb"
+      "maxSize": "1.3 Kb"
     },
     {
       "name": "apollo-link-batch",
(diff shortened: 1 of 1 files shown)


benjamn:
@JoviDeCroock Here's my latest attempt at whole-application bundle size measurement: https://github.com/apollographql/react-apollo/pull/2839

PR #969, opened 2019-03-05
Shrink apollo-link[-http-common] packages.

benjamn:
As https://github.com/apollographql/react-apollo/pull/2839 has helped to demonstrate, sharing dependencies between Apollo packages is a great way to improve overall bundle size, especially if the shared dependencies can be tree-shaken to remove unused exports.

This PR includes two changes to the `apollo-link` package, which is used directly or indirectly by any application that uses any of the packages in this repository:

1. Import `getOperationName` from `apollo-utilities`, since that implementation is almost certainly already imported by `apollo-client`.

2. Use `invariant.warn` and `InvariantError` instead of `console.warn` and `Error`, since those utilities allow long error strings to be stripped in production, thanks to https://github.com/apollographql/invariant-packages.

Together, these tweaks reduce the minified+gzip size of the `apollo-link` package from 1271 bytes to 1123 bytes, a 12% improvement.


rollup.config.js
```diff
@@ -70,6 +77,9 @@ export default name => [
           },
         },
       }),
+      invariantPlugin({
+        errorCodes: true,
```
benjamn: See https://github.com/apollographql/apollo-client/pull/4521 for more explanation of what this option does.

package.json
```diff
@@ -118,6 +118,7 @@
     "prettier": "1.15.2",
     "prettier-check": "2.0.0",
     "rollup": "1.2.3",
+    "rollup-plugin-invariant": "^0.4.2",
```
JoviDeCroock: I would remove the `^` since renovate makes a new PR for that otherwise

packages/apollo-link/package.json
```diff
@@ -34,6 +34,8 @@
     "watch": "tsc -w -p . & rollup -c -w"
   },
   "dependencies": {
+    "apollo-utilities": "^1.2.1",
```
JoviDeCroock: Same for these dependencies

rollup.config.js
JoviDeCroock: That's actually a really nice solution for having some errors in production. That rollup plugin is really smart

packages/apollo-link/package.json
benjamn: Hmm, I see the argument for `devDependencies`, but I'm a little worried about duplicating `apollo-utilities` in production bundles if we pin the exact version.

JoviDeCroock: Well, ofcourse it can't break because of a failing build since it's bundled in but well yes mostly up to how you want to handle these. I personally like having control over dependencies but that's my two cents on it.

benjamn: I agree application developers should use tools like `yarn.lock` or `package-lock.json` to pin all their dependencies, but from the perspective of this library, any version of `apollo-utilities` that exports `getOperationName` will work, which includes at least two years worth of past versions, and probably all future versions too.

JoviDeCroock: In that case I'm okay with it, I'll resolve this conversation

JoviDeCroock:
I'll probably try to find some low-hanging in link-packages today, so that might further reduce the bundlesize.

PR #970, opened 2019-03-05
chore: cut some low-hanging fruit in apollo-link

JoviDeCroock:
I wanted to do this package by package so we can accurately see what changes and how it affects bundle size, I hope this imposes no problem for anyone.

### DIFF
--- a/packages/apollo-link/src/link.ts
+++ b/packages/apollo-link/src/link.ts
@@ -16,24 +16,28 @@ import {
   createOperation,
 } from './linkUtils';
 
-const passthrough = (op, forward) => (forward ? forward(op) : Observable.of());
+function passthrough(op, forward) {
+  return forward ? forward(op) : Observable.of();
+}
 
-const toLink = (handler: RequestHandler | ApolloLink) =>
-  typeof handler === 'function' ? new ApolloLink(handler) : handler;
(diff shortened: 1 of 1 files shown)


packages/apollo-link/src/link.ts
```diff
@@ -33,10 +32,10 @@ export const from = (links: ApolloLink[]): ApolloLink => {
 export const split = (
   test: (op: Operation) => boolean,
   left: ApolloLink | RequestHandler,
-  right: ApolloLink | RequestHandler = new ApolloLink(passthrough),
+  right: ApolloLink | RequestHandler,
```
benjamn:
```suggestion
  right?: ApolloLink | RequestHandler,
```

packages/apollo-link/src/link.ts
```diff
@@ -102,9 +101,9 @@ export class ApolloLink {
   public split(
     test: (op: Operation) => boolean,
     left: ApolloLink | RequestHandler,
-    right: ApolloLink | RequestHandler = new ApolloLink(passthrough),
+    right: ApolloLink | RequestHandler,
```
benjamn:
```suggestion
    right?: ApolloLink | RequestHandler,
```

JoviDeCroock:
The change is quite minimal but might shave off 0.2KB in the total.

PR #984, opened 2019-03-13
Fixes the DedupLink shared observable unsubscribing early when shared

ms:
### Context

The DedupLink creates a real (single) observable to listen to when a
request is made and returns a shared observable to its caller. When a
second identical request is made, the real observable is left untouched
but the subscriber's callback functions is added to the list of
callbacks to be triggered when the request resolves and the shared
observable is returned.

### Problem

Unfortunately the shared observable returns cleanup logic that
unsubscribes from the real observable without checking whether only one
of many (shared) subscribers is unsubscribing. In the case where
DedupLink is in front of HttpLink, this leads to HttpLink aborting the
HTTP request while some callers are still waiting for a response.

### Fix

This change modifies the sharing code to use a Set of observers and only
unsubscribe from the underlying Observable when the last observer is
removed.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## vNEXT
 
+### apollo-link-dedup 1.0.17
+
+- Fixes an issue caused by the `DedupLink` shared observable returning 
+  cleanup logic that unsubscribes from the real observable, without 
+  checking whether only one of the many (shared) subscribers are 
+  unsubscribing. This caused problems when using `DedupLink` in front of 
+  `HttpLink`, as this lead to `HttpLink` aborting HTTP requests while some 
+  callers were still waiting for a response.  <br/>
+  [@ms](https://github.com/ms) in [#984](https://github.com/apollographql/apollo-link/pull/984)
+
+
 ## 2019-03-05
 
 ### General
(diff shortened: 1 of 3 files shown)


ms:
Just to add a bit more context around this: we were running into issues where requests were being made successfully in `cache-first` mode but failing when we switch to `network-only`. After a bit of debugging we realized that were were mounting, updating and unmounting the nested `Query` component multiple times (because of up-tree components resolving and updating their state, including an other `Query`) and that we happened to be sending the exact same query object and deep equal but non-identical variables object.

The result was that every query observable ended up with two subscribers: the one created in the `Query` component itself as well as the one created by `ObservableQuery.result()` after `Query.updateQuery`. Because the latter creates a `Promise` that unsubscribes on the next tick, we ran into race conditions where the `Query` component was unmounted and remounted between the original resolution and the `unsubscribe()`. When the second creation of the same `<Query />` triggered its fetch query, it was deduped by the `DedupLink` in the process (since the last observer hadn't unsubscribed yet) which was then aborted immediately thereafter when the `setTimeout` triggered.

A related question I have is whether it makes sense to use a `Promise` in `result()` in https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/ObservableQuery.ts#L121-L153 since the underlying `Observable` actually calls `next` multiple times (in `loading: true` and `loading: false` states, for example). It's possible I'm misunderstanding this code though

PR #985, opened 2019-03-14
Drop child package `package-lock.json` files

hwillson:
Since we're using lerna to manage and publish this monorepo, we don't need to maintain `package-lock.json` files in child packages. Maintaining a single `package-lock.json` in the root is enough (and child `package-lock.json` files are ignored at publish time by npm anyways).

References:
- https://docs.npmjs.com/files/package-lock.json
- https://github.com/lerna/lerna/issues/913


rtm:
I'm confused. I assumed without thinking about it that lerna was maintaining package-specific lock files. I recently cloned a repo that was about six months old and did lerna bootstrap and all hell broke loose as newer versions of things proved incompatible with other things. What is the reason again to prevent or not support the use of package-lock.json in packges, and it they are not supported, what is the workaround to lock down versions? How does a single package-lock.json at the top possibly suffice for managing versions within the packages, when the `npm install` in the package (invoked by `bootstrap`) doesn't use, or even see, the `package-lock.json` at the higher level?

PR #992, opened 2019-03-18
Avoid importing graphql/language/printer for getKey.

benjamn:
We have been slowly working to remove all uses of the graphql-js `print` function from Apollo client packages, since the printer weighs in at 4.86KB minified (without gzip); and, while that's not terribly huge, printing large queries takes precious time that we can usually avoid.

This particular usage is relatively easy to replace with `JSON.stringify`. This should be faster than actually printing the query, and we can serialize everything (`query`, `variables`, and `operationName`) with just one call to `JSON.stringify`, without manually joining the fragments with '|'.

It's reasonable to be concerned about the stability of `JSON.stringify`, since equivalent queries could have slightly different property ordering, but the official GraphQL printer doesn't make any guarantees about the stability or canonicality of the resulting string, either. The important thing is that `===` identical queries have the same string form, and that's definitely true here.

After this change, there is just [one remaining call to `print` in `apollo-link-http-common`](https://github.com/apollographql/apollo-link/blob/ed1800b4261566e91906cdb9cc45c7d004393175/packages/apollo-link-http-common/src/index.ts#L239), which seems a bit harder to replace, but at least you have the option of using something other than apollo-link-http if you really care about not reprinting the query before sending it to the server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 **Note:** This is a cumulative changelog that outlines all of the Apollo Link project child package changes that were bundled into a release on a specific day.
 
+## vNEXT
+
+### apollo-link
+
+- Avoid importing `graphql/language/printer` for `getKey`. <br/>
+  [@benjamn](https://github.com/benjamn) in [#992](https://github.com/apollographql/apollo-link/pull/992)
+
+
 ## 2019-03-14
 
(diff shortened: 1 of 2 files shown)


benjamn:
> 22 Bytes less in a package of this size is pretty good!

Agreed, though I definitely was not expecting that!

Issue #995, opened 2019-03-20
Handle rate limiting from RetryLink


mxstbr:
**Expected Behavior**

When a server returns a 429 rate limited response with a `Retry-After` header, the RetryLink does not retry until after the specified delay.

**Actual Behavior**

The RetryLink keeps retrying.

JoviDeCroock:
Thanks for the issue report, I'll take a look at it tonight.

mxstbr:
I tried that, unfortunately, the `error` does not contain the network status code.

JoviDeCroock:
Aha, this is indeed an issue we've been facing frequently. I'm going to try and look into this but it's quite hard to reproduce at this point. That's valuable information though thank you for that.

JoviDeCroock:
This feels like weird behavior to me:

https://codesandbox.io/s/ry49xr9nzm

Look at the logs, the statusCode seems to be contained in the networkError.

JoviDeCroock:
@mxstbr I have updated the sandbox to show the statusCode inside the retryLink aswell. If this does not fix your issue I can dig deeper into it but I'll need some more information.

mxstbr:
Yes exactly, I only want to retry after the server allows me to retry again.

JoviDeCroock:
Sorry for the confusion, I have been hunting for that statusCode for a while that it was drawing my attention of the main issue.

JoviDeCroock:
Hey @mxstbr if you would find the time could you check if this works for you. https://github.com/apollographql/apollo-link/pull/1004

Sorry about getting back to you this late

helfer:
@mxstbr I agree that the default retry link should check the `Retry-After` header. For now unfortunately there doesn't seem to be a way to work around it, because the delay function doesn't get the error (which should contain the status code btw, so if it doesn't, that's another issue).

PR #1004, opened 2019-03-29
feat: retry if server responds that

JoviDeCroock:
Fixes: https://github.com/apollographql/apollo-link/issues/995

This is a draft to see how everyone fares with this.


JoviDeCroock:
CircleCI crashed....

packages/apollo-link-retry/src/retryLink.ts
```diff
@@ -148,7 +148,21 @@ class RetryableOperation<TValue = any> {
...
     this.retryCount += 1;
-
+    // Retry if the server tells us to.
+    if (
+      error &&
+      error.response &&
+      error.response.status === 429 &&
+      error.response.headers['Retry-After']
+    ) {
+      this.scheduleRetry(
```
helfer: Doesn't this end up scheduling two retries if the response is a 429 (one in the `if` block, and one further down)?

helfer:
Hm, now I see that the delay function doesn't get the error, it only gets the count. It might be worth passing the error as a second argument to delay. I think that would be pretty reasonable, because one might legitimately want to retry after a different amount of time for different errors.

JoviDeCroock:
> Hm, now I see that the delay function doesn't get the error, it only gets the count. It might be worth passing the error as a second argument to delay. I think that would be pretty reasonable, because one might legitimately want to retry after a different amount of time for different errors.

We normally do pass it: https://github.com/apollographql/apollo-link/pull/1004/files#diff-cd9d084024219e2261760489591c11f2R174

Also given in typings of DelayFunction

PR #1028, opened 2019-04-23
Do not throw JSON parse error for not successful responses

maciekbb:
Right now if graphql server response is not successful and the response does not contain proper JSON (which I think is often the case) the following error is thrown `Unexpected token x in JSON at position 0` 

The issue was already described here https://github.com/apollographql/apollo-client/issues/2770

I believe that if response status is != 2xx the proper error to throw is `Response not successful: Received status code xyz` no matter what response body is.
TODO:

### DIFF
--- a/packages/apollo-link-http-common/src/__tests__/index.ts
+++ b/packages/apollo-link-http-common/src/__tests__/index.ts
@@ -28,7 +28,7 @@ describe('Common Http functions', () => {
     const operations = [createOperation({}, { query })];
 
     it('throws a parse error with a status code on unparsable response', done => {
-      const status = 400;
+      const status = 200;
       fetchMock.mock('begin:error', status);
       fetch('error')
         .then(parseAndCheckHttpResponse(operations))
--- a/packages/apollo-link-http-common/src/__tests__/index.ts
+++ b/packages/apollo-link-http-common/src/__tests__/index.ts
@@ -27,6 +27,22 @@ describe('Common Http functions', () => {
 
     const operations = [createOperation({}, { query })];
 
+    it('throws a server error with a status code on not successful response', done => {
+      const status = 400;
+      fetchMock.mock('begin:error', status);
+      fetch('error')
+        .then(parseAndCheckHttpResponse(operations))
+        .then(done.fail)
+        .catch(e => {
+          expect(e.statusCode).toBe(status);
+          expect(e.name).toBe('ServerError');
+          expect(e).toHaveProperty('response');
+          expect(e).toHaveProperty('result');
+          done();
+        })
+        .catch(done.fail);
+    });
+
     it('throws a parse error with a status code on unparsable response', done => {
       const status = 200;
       fetchMock.mock('begin:error', status);
--- a/packages/apollo-link-http/src/__tests__/sharedHttpTests.ts
+++ b/packages/apollo-link-http/src/__tests__/sharedHttpTests.ts
@@ -842,9 +842,30 @@ export const sharedHttpTest = (
 
     const body = '{';
     const unparsableJson = jest.fn(() => Promise.resolve(body));
-    it('throws an error if response is unparsable', done => {
+    it('throws an error if response is unparsable and status >= 300', done => {
       fetch.mockReturnValueOnce(
-        Promise.resolve({ status: 400, text: unparsableJson }),
+        Promise.resolve({ status: 503, text: () => Promise.resolve('foo') }),
+      );
+
+      const link = createLink({ uri: 'data', fetch });
+
+      execute(link, { query: sampleQuery }).subscribe(
+        result => {
+          done.fail('should not have called result');
+        },
+        e => {
+          expect(e.name).toEqual('ServerError');
+          expect(e.message).toMatch(
+            /Response not successful: Received status code 503/,
+          );
+          expect(e.statusCode).toBe(503);
+          done();
+        },
+      );
+    });
+    it('throws an error if response is unparsable and status is < 300', done => {
+      fetch.mockReturnValueOnce(
+        Promise.resolve({ status: 200, text: unparsableJson }),
       );
       const link = createLink({ uri: 'data', fetch });
 
@@ -854,7 +875,7 @@ export const sharedHttpTest = (
         },
         makeCallback(done, e => {
           expect(e.message).toMatch(/JSON/);
-          expect(e.statusCode).toBe(400);
+          expect(e.statusCode).toBe(200);
           expect(e.response).toBeDefined();
           expect(e.bodyText).toBe(body);
         }),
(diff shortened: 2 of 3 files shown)


packages/apollo-link-http-common/src/__tests__/index.ts
```diff
@@ -28,7 +28,7 @@ describe('Common Http functions', () => {
     const operations = [createOperation({}, { query })];
 
     it('throws a parse error with a status code on unparsable response', done => {
-      const status = 400;
```
JoviDeCroock: Why does this need changing this should not have effect from the PR

packages/apollo-link-http/src/__tests__/sharedHttpTests.ts
```diff
@@ -842,9 +842,30 @@ export const sharedHttpTest = (
 
     const body = '{';
     const unparsableJson = jest.fn(() => Promise.resolve(body));
-    it('throws an error if response is unparsable', done => {
+    it('throws an error if response is unparsable and status >= 300', done => {
```
JoviDeCroock: >= 500? your PR sais 5xx responses

packages/apollo-link-http/src/__tests__/sharedHttpTests.ts
```diff
@@ -842,9 +842,30 @@ export const sharedHttpTest = (
...
+          expect(e.name).toEqual('ServerError');
+          expect(e.message).toMatch(
+            /Response not successful: Received status code 503/,
+          );
+          expect(e.statusCode).toBe(503);
+          done();
+        },
+      );
+    });
+    it('throws an error if response is unparsable and status is < 300', done => {
```
JoviDeCroock: < 500

packages/apollo-link-http-common/src/__tests__/index.ts
maciekbb: Well if status code is 400 I expect it to be `ServerError` instead of `ServerParseError`. The reason is that if status code is not ok I cannot assume the response is valid JSON.

maciekbb: I will add an additional unit to cover this case

packages/apollo-link-http/src/__tests__/sharedHttpTests.ts
maciekbb: Well in the codebase I noticed there is only distinction between status codes 2xx and others (not successful). I will change the PR description.

packages/apollo-link-http/src/__tests__/sharedHttpTests.ts
maciekbb: Actually this case covers successful responses (so status 2xx)

If response is successful and body is not valid (parsable) JSON than I actually expect `ServerParseError`

maciekbb:
@JoviDeCroock thanks for the review! I updated the PR description and added a test case.

JoviDeCroock:
Honestly, your server not responding erroneous responses as JSON feels more like a server setup mistake than anything else.
Correct me if I’m wrong but just returning a string does not happen often, this reduces detail in errors.

maciekbb:
@JoviDeCroock The request might not even make it to graphql server but get lost in a different layer of a system (like authentication/authorisation) or the service might not be available (503 response). True, it might be setup mistake, however at the same time from client perspective I'd rather receive `ServerError` than mysterious `Unexpected token x in JSON at position y`

JoviDeCroock:
Yes, that is why I'm saying to justify not parsing it by looking at the headers instead of just not parsing it at all.

helfer:
I agree with @JoviDeCroock. I believe the HTTP link only puts `application/json` in the accept header of the request. If the server responds with anything other than JSON (or maybe a 406, if you want to get really fancy), that's really a server issue.

The real problem with returning a ServerError when the response body couldn't be parsed, is that it would mask the fact that the response couldn't be parsed. For that reason, the current ServerParseError contains all the information from the response that ServerError includes, which will let you work around the issue (see the [type definitions](https://github.com/apollographql/apollo-link/blob/ef1c2d182fd6e437a60278e9b2108539b7a8c6fd/packages/apollo-link-http-common/src/index.ts#L15)).

@maciekbb ServerParseError still includes the status code and the response body, so if you want you can put a link in your chain that transforms the ServerParseError to whatever you want in your application. That would be an alternative workaround to your PR that doesn't change existing behavior. If you control the server response however, I suggest you go in and fix that so that it returns json as it should according to the HTTP protocol.

Suggestion: To improve the user experience, one thing worth considering might be to check the content type of the response if parsing fails, and to change the error message to say that the server didn't respond with JSON, and maybe a helpful suggestion for how to fix it, or a link to docs that explain how to fix / work around the issue.

maciekbb:
Thanks for your opinion! I will probably go with custom apollo link, like this 

```
import {ApolloLink, Observable} from 'apollo-link';

export const handleParseErrorLink = new ApolloLink((operation, forward) => {
  return new Observable(observer => {
    const sub = forward(operation).subscribe({
      next: observer.next.bind(observer),
      error: err => {
        if (err instanceof SyntaxError && err.statusCode >= 300) {
          const mappedError = new Error(
            `Response not successful: Received status code ${err.statusCode}`
          );

          observer.error(mappedError);
        }

        return observer.error(err);
      },
      complete: observer.complete.bind(observer),
    });

    return () => {
      if (sub) sub.unsubscribe();
    };
  });
});
```

jscheid:
> If you control the server response however, I suggest you go in and fix that so that it returns json as it should according to the HTTP protocol.

"Should" being the operative word here, the server is not at all required to honor the `Accept` header as per RFC 7231 section 5.3.2 (emphasis mine):

> If the header field is present in a request and none of the available representations for the response have a media type that is listed as acceptable, the origin server can either honor the header field by sending a 406 (Not Acceptable) response _or disregard the header field by treating the response as if it is not subject to content negotiation._

I'm pretty sure even widely used components/services such as Nginx, [CloudFlare](https://blog.apollographql.com/caching-graphql-results-in-your-cdn-54299832b8e2) etc. don't always know how to return errors in JSON format and will return plain text at least some of the time -- and doing so is perfectly compliant with the spec.

Also, don't forget that this is not just about a handful of proxies or load balancers. There are probably 10,000+ different pieces of circuitry that might sit between the client and server. For example, there are [over a thousand pieces of middleware for Rails](https://rubygems.org/search?utf8=%E2%9C%93&query=rack+middleware) alone, all of which can short-circuit the request and return an error response for all sorts of reasons: say, rate limiting. And I'd wager more often than not they don't even look at the `Accept` header.

What's wrong with simply only parsing the response as JSON when `Content-Type: application/json`? If a server flags an error response as JSON when it isn't (or, arguably, vice versa, if it forgets to set `Content-Type`) then that is indeed a bug and should be fixed upstream.

PR #1041, opened 2019-05-09
Changed markdown to html to render img

hyxos:
Putting markdown inside an html tag doesn't parse properly so the images don't show up. Putting the source it inside an image tag should fix it. I can't see the preview so a width or height may need to be added for proper scaling.

This is purely a cosmetic change for the documents so shouldn't effect the code base


hyxos:
No problem, glad to help  😁

Issue #541, opened 2018-03-09
Apollo-link-retry: Doesn’t support retry on GQL errors only network errors


RylanH:
I’d like to see the retry link allow for retries on graphQL errors, not just network ones. Sometimes there are transient 5xx style errors in a GQL resolver.

pthrasher:
@RylanH The below link "promotes" graphql errors to real errors and retry link will process them. You must put this link between your http link and your retry link.


```javascript
      new ApolloLink((operation, forward) => {
        return forward(operation).map((data) => {
          if (data && data.errors && data.errors.length > 0) {
            throw new Error('GraphQL Operational Error');
          }
          return data;
        });
      }),
```

PR #1067, opened 2019-06-03
feat:(apollo-link-retry): Handle retries for GraphQL errors in the response

astorije:
Resolves https://github.com/apollographql/apollo-link/issues/541

TODO:

Changes at lines 97 and 102 in `retryLink.ts` are obviously wrong. When I do not disable handling `onComplete`, the query does not get retried because the observable calls `complete` even though the response contains an error.
I just disabled these to test and the retry logic works exactly as expected (for POC) but I'm a little blocked on fixing the `complete` handling.

Anyone would care to help me a little? I have "Allow edits from maintainers" on, if so. Thank you in advance, we would _love_ to get this feature 🙏

### DIFF
--- a/packages/apollo-link-retry/README.md
+++ b/packages/apollo-link-retry/README.md
@@ -3,7 +3,7 @@ title: apollo-link-retry
 description: Attempt an operation multiple times if it fails due to network or server errors.
 ---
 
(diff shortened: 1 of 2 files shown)


himat:
Hello, was this never finished?
What is the proper way to make graphql retry the query if it failed due to a graphql error?

PR #1095, opened 2019-06-19
Remove reference to GlobalFetch in Typescript

### LINKED ISSUE #1094: apollo-link-http-common references type GlobalFetch, to be removed in Typescript 3.6
**Expected Behavior**
No error when compiling [Definitely Typed types for apollo-upload-client](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/apollo-upload-client/index.d.ts):

```ts
import { ApolloLink } from "apollo-link";
import { HttpOptions } from "apollo-link-http-common";
export { ReactNativeFile } from "extract-files";
export function createUploadLink(linkOptions?: HttpOptions): ApolloLink;
```

**Actual Behavior**
Error in apollo-link-http-common's types, which refer to `GlobalFetch`. GlobalFetch will be removed in Typescript 3.6 in favour of [WindowOrWebWorker, which has the same `fetch` method](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch):

```ts
export interface HttpOptions {
    // ........
    /**
     * A `fetch`-compatible API to use when making requests.
     */
    fetch?: GlobalFetch['fetch'];
    // .........
}
```

**A _simple_ reproduction**
```sh
npm install -g typescript@next
cd packages/apollo-link-http-common/src
tsc
```

Gives the errors:

```
index.ts:74:11 - error TS2304: Cannot find name 'GlobalFetch'.

74   fetch?: GlobalFetch['fetch'];
             ~~~~~~~~~~~

index.ts:174:39 - error TS2304: Cannot find name 'GlobalFetch'.

174 export const checkFetcher = (fetcher: GlobalFetch['fetch']) => {
```

(This is not the right way to build this file, so there are 4 other errors. You will only see those 4 if you use `typescript@latest`)

The fix is to replace `GlobalFetch` with `WindowOrWorkerGlobalScope`. I'll put up a PR that does this shortly.

sandersn:
Fixes compilation of apollo-link-http-common in the upcoming Typescript 3.6.

Fixes #1094

There is no new test for this -- if it builds then it works with whatever version of Typescript is on CI; if apollo-link doesn't already test with multiple versions of TS then I don't think this is the place to make that happen.

### DIFF
--- a/packages/apollo-link-http-common/src/index.ts
+++ b/packages/apollo-link-http-common/src/index.ts
@@ -71,7 +71,7 @@ export interface HttpOptions {
   /**
    * A `fetch`-compatible API to use when making requests.
    */
-  fetch?: GlobalFetch['fetch'];
+  fetch?: WindowOrWorkerGlobalScope['fetch'];
 
   /**
    * An object representing values to be sent as headers on the request.
@@ -171,7 +171,7 @@ export const parseAndCheckHttpResponse = operations => (response: Response) => {
   );
 };
 
-export const checkFetcher = (fetcher: GlobalFetch['fetch']) => {
+export const checkFetcher = (fetcher: WindowOrWorkerGlobalScope['fetch']) => {
   if (!fetcher && typeof fetch === 'undefined') {
     let library: string = 'unfetch';
     if (typeof window === 'undefined') library = 'node-fetch';


sandersn:
@benjamn It doesn't look like all the circleci tasks are running. Is it backlogged or something? The build succeeded but Filesize, Monorepo, Prettier, etc have not started.

akleiner2:
Hey @benjamn @hwillson! Would it be possible to do a release that includes this commit? We're looking at upgrading to TS3.6 ourselves, and would love to have this as a part of the upgrade :).

sarink:
^ 🙏 

In the meantime... Is there any workaround?

smyrick:
@sarink @akleiner2 The latest release fixed the issue for us

PR #1113, opened 2019-07-15
Fix broken links to error in http docs

icco:
Just small doc fixes seen from browsing https://www.apollographql.com/docs/link/links/http/


JoviDeCroock (approved):
Looks good, thanks for taking your time to correct this.

PR #1126, opened 2019-08-14
Use `forEach` instead of `map` for error logging

DaleSeo:
I think the more proper way would be using `forEach` here, as `map` is needlessly returning an array of undefined elements.


DaleSeo:
@benjamn Do we have any reason to keep this PR open for this long? Please let me know if there is anything for me to do to get this merged. Thank you!

PR #1196, opened 2019-11-11
Adds missing peer dependency

arcanis:
`apollo-link-error` has a transitive peer dependency on `graphql` through `apollo-link`, and must be marked as such. Cf my article why [implicit transitive peer dependencies are bad](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0) 🙂


cometkim:
And there are few more; apollo-link-context also doesn't have peerDeps.

I can't use apollo with yarn 2 because of this. Any plan to update?

Issue #1168, opened 2019-10-17
Ignoring errors does not work as in the docs when using Typescript


syeo66:
**Expected Behavior**
According to https://www.apollographql.com/docs/react/data/error-handling/#ignoring-errors I should be able to modify `response.errors.

**Actual Behavior**
```
Type 'GraphQLError[] | null' is not assignable to type 'readonly GraphQLError[] | undefined'.
  Type 'null' is not assignable to type 'readonly GraphQLError[] | undefined'.
```

**A _simple_ reproduction**
Try the code at https://www.apollographql.com/docs/link/links/error/#ignoring-errors with typescript

apollo-link-error@1.1.12

PR #1240, opened 2020-01-21
Update link-error docs to reference `response.errors = undefined`

dangreenisrael:
The docs incorrectly show that `response.error` can be null. However in the graphql-js package that defines `ExecutiionResult` in is is either an array of `GraphQLError` or undefined. Null is not a valid option.

https://github.com/graphql/graphql-js/blob/master/src/execution/execute.d.ts#L52

Issue reference: https://github.com/apollographql/apollo-link/issues/1168

TODO:

### DIFF
--- a/packages/apollo-link-error/README.md
+++ b/packages/apollo-link-error/README.md
@@ -81,12 +81,12 @@ One caveat is that the errors from the new response from retrying the request do
 
 ## Ignoring errors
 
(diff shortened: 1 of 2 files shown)


dangreenisrael:
I didn't update the changelog as its a docs update, but I'd be happy to do it if you like 🙂

PR #1249, opened 2020-02-19
Add warnings about observers to the documentation

nsnyder:
This is intended to address the concerns in Issue #298. It adds warnings about using subscriptions on observers and suggests alternate methods of accomplishing the desired goal.


nsnyder:
@evans Would it be possible to get this merged into the documentation until a code fix exists for issue #298 ?

Issue #1261, opened 2020-04-04
Type error after upgrading to graphql@15


tyler-johnson:
I upgraded to graphql.js 15.0.0 and now I have this type error:

```
../../node_modules/apollo-link/lib/types.d.ts(23,56): error TS2315: Type 'ExecutionResult' is not generic.
```

ceefour:
I also got this issue while using it with NestJS:

```
D:\project_amanah\soluvas\soluvas-platform-nest>npm run start:dev

> soluvas-platform-nest@0.0.1 start:dev D:\project_amanah\soluvas\soluvas-platform-nest
> nest start --watch

[03:41:16] Starting compilation in watch mode...

node_modules/apollo-link/lib/types.d.ts:23:56 - error TS2315: Type 'ExecutionResult' is not generic.

23 }, C = Record<string, any>, E = Record<string, any>> = ExecutionResult<TData> & {
                                                          ~~~~~~~~~~~~~~~~~~~~~~

[03:41:32] Found 1 error. Watching for file changes.
```

package.json snippet:

```json
  "dependencies": {
    "@nestjs/common": "^7.0.0",
    "@nestjs/core": "^7.0.0",
    "@nestjs/graphql": "^7.2.0",
    "@nestjs/platform-fastify": "^7.0.7",
    "@nestjs/typeorm": "^7.0.0",
    "apollo-server-fastify": "^2.11.0",
    "graphql": "^15.0.0",
    "graphql-tools": "^4.0.7",
    "mysql": "^2.18.1",
    "reflect-metadata": "^0.1.13",
    "rimraf": "^3.0.2",
    "rxjs": "^6.5.4",
    "typeorm": "^0.2.24"
  },
  "devDependencies": {
    "@nestjs/cli": "^7.0.0",
    "@nestjs/schematics": "^7.0.0",
    "@nestjs/testing": "^7.0.0",
    "@types/express": "^4.17.3",
    "@types/jest": "25.1.4",
    "@types/node": "^13.9.1",
    "@types/supertest": "^2.0.8",
    "@typescript-eslint/eslint-plugin": "^2.23.0",
    "@typescript-eslint/parser": "^2.23.0",
    "eslint": "^6.8.0",
    "eslint-config-prettier": "^6.10.0",
    "eslint-plugin-import": "^2.20.1",
    "jest": "^25.1.0",
    "prettier": "^1.19.1",
    "supertest": "^4.0.2",
    "ts-jest": "25.2.1",
[...]

PR #1262, opened 2020-04-06
Remove TData generic, fixing GraphQL 15 compile

zackdotcomputer:
In earlier versions of GraphQL, the ExecutionResult took a generic parameter, which it then typecast its return type as. We were, however, just sending `{ [key: string]: any }` for this parameter in all internal cases.

As of GraphQL 15, that generic parameter has been removed, (see this PR)[https://github.com/graphql/graphql-js/pull/2490/commits], and the return type has been fixed to always be `{ [key: string]: any }`.

Since this is already compatible with our default, this PR removes the generic parameter to bring our typing in line with GraphQL's typing. This is, however, a breaking change for anyone who was depending on this generic parameter to do their typecast for them.

TODO:

### DIFF
--- a/packages/apollo-link/CHANGELOG.md
+++ b/packages/apollo-link/CHANGELOG.md
@@ -9,7 +9,8 @@ the top level [`CHANGELOG.md`](https://github.com/apollographql/apollo-link/blob
 
 ### 1.2.4
 
-- No changes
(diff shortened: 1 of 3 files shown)


zackdotcomputer:
Signed CLA, updated changelog. This fixes #1261 .

hwillson:
Thanks for working on this @zackdotcomputer. We've decided to merge https://github.com/apollographql/apollo-link/pull/1263 to address this instead, since it's backwards compatible. Thanks again!

PR #1263, opened 2020-04-07
Support GraphQL v15

kamilkisiela:
ExecutionResult no longer has generic type so we use our own ExecutionResult extended by original to keep the shape of result backwards compatible


kamilkisiela:
#1261

hwillson:
Thanks very much for working on this @kamilkisiela!

PR #1269, opened 2020-04-30
feat: introduce batching mode

dora-gt:
Thinking of this change.

Introduce BatchingMode which enables users to expressly specify the timing of executing the batched operations.

Feedbacks welcome.


dora-gt:
I'd like to this PR to be merged, what do you think?
Do you guys think this is not necessary or should I do something?

dora-gt:
ping @hwillson @benjamn

PR #1275, opened 2020-05-20
Update apollo-utilities to 1.3.4

onlined:
Before 1.3.4, `apollo-utilities` didn't have `graphql@^15.0.0` as a
peer dependency, so when `apollo-link` is already installed and
user switched to `graphql@15.0.0` they could have peer dependency
warnings. This update solves the problem.

TODO:


agataHappy:
Bump

